PR #26, opened 2018-11-21
Add more questions

sastels:
Add final-form and the questions for the prototype


frontend/src/Screen2.js
```diff
@@ -0,0 +1,80 @@
...
+
+const submitButton = css`
+  margin-top: 20pt;
+`
+
+const onSubmit = () => {
+  navigate('form3')
+}
+
+const validate = () => {}
```
sleepycat: Placeholder?

frontend/src/Screen1.js
```diff
@@ -0,0 +1,84 @@
...
+        >
+          <Trans>Submit</Trans>
+        </Button>
+      </form>
+    )}
+  />
+)
+
+export class Screen1 extends Component {
+  render() {
```
sleepycat: Generally when you aren't using [React lifecycle methods](https://reactjs.org/docs/react-component.html#the-component-lifecycle) components should be written as [stateless functional components](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc). There are [varying levels of excitement about the "stateless functional"](https://twitter.com/jamiebuilds/status/1057313713835794432) parts but it's roundly considered a good idea. The recently announced [hooks](https://reactjs.org/docs/hooks-intro.html) are going to further reduce the appeal of class based components.

frontend/src/__tests__/LandingPage.test.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import { LandingPage } from '../LandingPage'
+
+describe('LandingPage', () => {
+  it('renders', () => {
+    mount(<LandingPage />)
```
sleepycat: `mount` returns a wrapper which would be truthy, so the test should pass, as long as the component is properly exported and renders. I guess at this early stage "it can be imported and successfully renders anything at all" is a reasonable test. [`shallow`](https://stackoverflow.com/a/38747914/130006) might be a faster way to accomplish the same task.

frontend/src/Screen1.js
sastels: this is good to know 👍

PR #34, opened 2018-11-23
Hook front end up to api / arango

sastels:
actually save the data from the frontend to arango


api/index.js
```diff
@@ -3,9 +3,9 @@ const { db } = require('./src/db')
 
 Server(db)
   .then(server => {
-    server.listen({ port: 3000 }, () =>
+    server.listen({ port: 4000 }, () =>
       // eslint-disable-next-line no-console
-      console.log(`🚀 API listening on port 3000`),
+      console.log(`🚀 API listening on port 4000`),
```
sleepycat: This is going to break the Kubernetes config. If we need to do this could we pass in something from the ENV?

sastels:
🎉 It's working!

PR #49, opened 2018-11-29
Use docker-compose for local dev

sleepycat:
This commit adds a docker-compose file to bring up a working development environment.
I've added a make file to simplify things a little further, which leaves us with two simple commands to get things going:

```sh
make setup
DB_USER=root DB_PASSWORD=test make dev
```

This will bring up the frontend listening on localhost:3000, the API, at
localhost:3000/graphql and ArangoDB, with it's admin interface on
localhost:8529.

The first time you run this you will need to log into arangodb and create a database (called cybercrime) and a collection (called reports).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-iHELL ?= /usr/bin/bash
+SHELL ?= /usr/bin/bash
 # set up minikube for local testing.
 setup:
 	docker-compose build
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,7 @@
+iHELL ?= /usr/bin/bash
```
sastels: to iHell with you

PR #60, opened 2018-12-03
Add pageview tracking via Google Analytics

sleepycat:
Just what it says on the tin. This adds a `<TrackPageViews />` component.


frontend/src/Screen1.js
```diff
@@ -85,6 +86,7 @@ export const Screen1 = () => (
       <Trans>Describe what happened</Trans>
     </H1>
 
+    <TrackPageViews />
```
sastels: very clean 💯

PR #71, opened 2018-12-12
Extract Stats from Thanks component

sleepycat:
The Thanks component was getting unwieldy. This commit extracts a Stats
component from it, as well as some other functions and adds proper testing for
the Stats component.


frontend/src/__tests__/Stats.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+      let wrapper = mount(
+        <MockedProvider mocks={mocks} addTypename={false}>
+          <I18nProvider language={'en'} catalogs={catalogs}>
+            <Stats />
+          </I18nProvider>
+        </MockedProvider>,
+      )
+
+      await wait(0)
```
sastels: wait(0)?

sleepycat: If you throw some console.log statements into the "function as a child" that we are passing to the Query component you can see it's being called multiple times. This lets the promise that the client returns resolve so our Query component code gets called again with the actual data.

PR #77, opened 2019-01-07
Split WhatHappenedForm out of Screen1

sleepycat:
This commit got bigger than anticipated, but the idea here is splitting Screen1 into some smaller pieces.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -26,7 +26,6 @@
     "@lingui/react": "^2.7.2",
     "@reach/component-component": "^0.1.1",
     "@reach/router": "^1.2.1",
-    "acorn": "^6.0.4",
     "apollo-cache-inmemory": "^1.3.11",
     "apollo-client": "^2.4.7",
     "apollo-link-error": "^1.1.2",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -7,6 +7,7 @@
     "dev": "razzle start",
     "build": "razzle build",
     "heroku-postbuild": "razzle build",
+    "dbg": "node --inspect-brk node_modules/.bin/jest --runInBand",
     "test": "jest",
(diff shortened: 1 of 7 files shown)


frontend/package.json
```diff
@@ -7,6 +7,7 @@
     "dev": "razzle start",
     "build": "razzle build",
     "heroku-postbuild": "razzle build",
+    "dbg": "node --inspect-brk node_modules/.bin/jest --runInBand",
```
sastels: what does this do?

frontend/package.json
```diff
@@ -25,6 +26,7 @@
     "@lingui/react": "^2.7.2",
     "@reach/component-component": "^0.1.1",
     "@reach/router": "^1.2.1",
+    "acorn": "^6.0.4",
```
sastels: Are we using this? I can't find it elsewhere in the code.

frontend/src/WhatHappenedForm.js
```diff
@@ -0,0 +1,51 @@
...
+const validate = () => {}
+
+export const WhatHappenedForm = ({ onSubmit }) => (
+  <Form
+    onSubmit={onSubmit}
+    validate={validate}
+    render={({ handleSubmit, pristine, invalid }) => (
+      <form onSubmit={handleSubmit}>
+        <div>
+          <label htmlFor="whatHappened">What Happened</label>
```
sastels: So we didn't used to have this label. I assume without it we're damaging accessibility?

frontend/package.json
sleepycat: This is a handy shortcut for using the node debugger.

frontend/package.json
sleepycat: I can't remember why that was in there. I've pushed another commit that removes it.

PR #80, opened 2019-01-07
Change LanguageSwitcher to a button

### LINKED ISSUE #63: Language toggle
Manual testing with a keyboard + voice recognition I am unable to active the language element. 
A screen reader user may be able to access this by pulling up a list of links/buttons on the page, but a keyboard user will be unable to do so.

sleepycat:
This commit changes the LanguageSwitcher component to output a button. The
previous "a" element had no href attribute (since it wasn't navigating
anywhere) and was therefore not included in the accessibility tree. Switching
to a button fixes this and closes #63

### DIFF
--- a/frontend/src/LanguageSwitcher.js
+++ b/frontend/src/LanguageSwitcher.js
@@ -1,33 +1,41 @@
 import React from 'react'
 import { Query, Mutation } from 'react-apollo'
-import styled, { css } from 'react-emotion'
+import { css } from 'react-emotion'
 import {
   GET_LANGUAGE_QUERY,
   CHANGE_LANGUAGE_MUTATION,
 } from './utils/queriesAndMutations'
 
-const A = styled('a')`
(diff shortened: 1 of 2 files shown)


frontend/src/LanguageSwitcher.js
```diff
@@ -1,33 +1,41 @@
...
 import { Query, Mutation } from 'react-apollo'
-import styled, { css } from 'react-emotion'
+import { css } from 'react-emotion'
 import {
   GET_LANGUAGE_QUERY,
   CHANGE_LANGUAGE_MUTATION,
 } from './utils/queriesAndMutations'
 
-const A = styled('a')`
+const buttonThatLooksLikeALink = css`
```
sastels: I think we were / are just using default font sizes here, it's smaller than it used to be.
old:
image

new
image

sleepycat: Good catch! Fixed!

PR #91, opened 2019-01-17
Basic PWA

sastels:
Basic PWA.
- starts a service worker and caches page for offline use
- can be installed on the homescreen of a mobile device and run without browser controls/bookmarks/etc


sastels:
Also, @evadb @sleepycat our CI isn't running ☝️

frontend/public/service-worker.js
```diff
@@ -0,0 +1,38 @@
+const filesToCache = ['/']
+
+const staticCacheName = 'pages-cache-v1'
+
+// Install the files we've identified
+self.addEventListener('install', event => {
```
sleepycat: self? Well, how about that, service workers get their own special [global object](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Functions_and_classes_available_to_workers). Interesting!

sastels:
[Add basic PWA setup](https://trello.com/c/Uj9VkjnW/81-add-basic-pwa-setup)

PR #94, opened 2019-01-24
Tweak PWA

sastels:
Problem: the bundle file has a random hash in it due to Razzle, so the precaching on install fails.

In this PR the service worker  fetches Razzle’s asset file and uses that to get the file name of the bundle.


sastels:
[Add basic PWA setup](https://trello.com/c/Uj9VkjnW/81-add-basic-pwa-setup)

PR #108, opened 2019-02-07
replace console.log with Logger

sastels:
Use `@cdssnc/logdriver` for logging

### DIFF
--- a/frontend/public/service-worker.js
+++ b/frontend/public/service-worker.js
@@ -40,7 +40,6 @@ self.addEventListener('activate', event => {
 
 // Network falling back to the cache
 self.addEventListener('fetch', function(event) {
-  console.log('Fetch', event.request)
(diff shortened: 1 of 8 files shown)


frontend/public/service-worker.js
```diff
@@ -40,7 +40,6 @@ self.addEventListener('activate', event => {
 
 // Network falling back to the cache
 self.addEventListener('fetch', function(event) {
-  console.log('Fetch', event.request)
```
evadb: is this one not also being replaced by a Logger call?

sastels: nah, that was a leftover log from when I was getting the PWA code working, slipped through. I meant to comment on that in the PR but forgot!

evadb: 👍

PR #111, opened 2019-02-07
Remove unused variable to fix the build

sleepycat:


frontend/src/server.js
```diff
@@ -4,7 +4,6 @@ import express from 'express'
 import requestLanguage from 'express-request-language'
 import { renderToString } from 'react-dom/server'
 import { ServerLocation } from '@reach/router'
-import { Logger } from '@cdssnc/logdriver'
```
sastels: Wait, aren’t we still using Logger down near line 93?

sastels: yeah, I get an error in the console
image

sastels: Ah, same error in master. Basically, the issue is that in that code `Logger` is being used in a script being sent to the client, not run on the server where the import is happening. So I think we want to not use Logger in that script (and delete the import as you already have)

PR #118, opened 2019-02-12
Add Paragraph component

evadb:
It BEGINS. This PR adds the paragraph component, presets, test & an example usage in the code.

### DIFF
--- a/frontend/src/__tests__/Paragraph.test.js
+++ b/frontend/src/__tests__/Paragraph.test.js
@@ -1,0 +1,33 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { ThemeProvider } from 'emotion-theming'
+import theme from '../theme'
+import { Paragraph, P, Lead } from '../components/paragraph'
+
+describe('Paragraph', () => {
+  const example = 'example'
+  const wrapper = (
+    <ThemeProvider theme={theme}>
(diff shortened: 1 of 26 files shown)


frontend/src/__tests__/Paragraph.test.js
```diff
@@ -0,0 +1,31 @@
...
+  const example = 'example'
+  const wrapper = (
+    <ThemeProvider theme={theme}>
+      <Paragraph color="black">{example}</Paragraph>
+    </ThemeProvider>
+  )
+
+  it('Renders a Paragraph component, P and Lead presets without crashing', () => {
+    const div = document.createElement('div')
+    ReactDOM.render(wrapper, div)
```
sastels: Why not use enzyme, for example as in https://github.com/cds-snc/report-a-cybercrime/blob/master/frontend/src/__tests__/Home.test.js#L14-L20

sastels:
Wait, do you want to be merging into `master` instead of `add-components-2`?

image

frontend/src/__tests__/Paragraph.test.js
evadb: Largely because I stole how govuk react does there tests so I could move more quickly, I have no strong feelings on how we test! I'm actually doing some diving into how to test things today

PR #120, opened 2019-02-13
Use new components for headers

sastels:
also deleted the old `utils/headers.js` file that contained the old headers.


frontend/src/Screen1.js
```diff
@@ -37,7 +37,7 @@ export const Screen1 = () => (
       </Link>
     </Breadcrumb> */}
     <div>Placeholder for Breacrumb</div>
-    <H1>
+    <H1 fontSize={[5, null, 7]}>
```
evadb: I should have changed this before pushing the PR, but when I added in the Headers the preset font size is [6, null, 7], but as it turns out 6 is waaaaay too large for mobile. What would be nice is if you could update the preset for H1 in presets.js? That was we can just call H1 instead of customizing the font size every time

sastels: 👍

JuliannaR:
Looks good. Thanks!

PR #121, opened 2019-02-13
Use new paragraph component

sastels:
use new `<P>` instead of `<p>`


frontend/src/theme/index.js
```diff
@@ -39,6 +39,7 @@ const theme = {
     lightBlue: '#2b8cc4',
     purple: '#4c2c92',
     yellow: '#ffbf47',
+    crimson: '#dc143c',
```
evadb: niiiiiice!!

PR #122, opened 2019-02-13
Add unordered lists, ordered lists & list items

evadb:
This PR adds components for unordered lists, ordered lists & list items! As per the usual, added an example on the landing page but left everything else untouched.

### DIFF
--- a/frontend/src/components/unordered-list/presets.js
+++ b/frontend/src/components/unordered-list/presets.js
@@ -5,8 +5,8 @@ import { css } from '@emotion/core'
 
 export const Ul = props => (
   <UnorderedList
-    fontSize={[1, null, 2]}
-    lineHeight={[1, null, 2]}
+    fontSize={[2, null, 3]}
+    lineHeight={[2, null, 3]}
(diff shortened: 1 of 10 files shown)


frontend/src/components/unordered-list/presets.js
```diff
@@ -0,0 +1,57 @@
...
+  children: PropTypes.any,
+}
+
+export const UlNone = props => (
+  <UnorderedList
+    fontSize={[2, null, 3]}
+    lineHeight={[2, null, 3]}
+    mb={4}
+    css={css`
+      list-style-type: none;
```
sastels: why not just have `Ul` and have the `list-style-type` customizable?

evadb: Could definitely be done, but would require some more logic in the UnorderedList component since styled-components doesn't gift us with a prop to handle it (sad). I have a vague recollection of trying, hitting a road block, and then doing this instead. Definitely open to refactoring after this PR.

Oh, also, because UlNone would require it's own preset without margins anyway. Hmm 🤔

PR #123, opened 2019-02-13
Upgrade packages

sastels:
upgraded a bunch of packages in `frontend`. Remaining is `jest` and `babel-jest`, which have major upgrades that broke things.


maxneuvians:
Interesting bundle size only increased by 1 byte

sastels:
bundle now 200B bigger

maxneuvians:
Byte math seems to be off, probably rounding error

evadb:
Can we upgrade React too in this one? :o

sastels:
@evadb AFAIK react was already upgraded at some point? (Latest is 16.8.1 as is in the lock file)
https://github.com/cds-snc/report-a-cybercrime/blob/upgrade_packages/frontend/package-lock.json#L3440

PR #124, opened 2019-02-13
Refactor lists to accept listStyleType prop

evadb:
@sastels made a really good point that being able to pass listStyleType as a prop would be much nicer syntax, so... here it is! Now both Ordered and Unordered lists will accept any legitimate list style type (https://www.w3schools.com/cssref/pr_list-style-type.asp) as a prop, but default to number and disc respectively (since these are by far the most used types. Passing in none results in a special list with no left padding. 

Also I tested ordered lists on the landing page and realized.... I really like it as an ordered list, since it's a step by step list! So I left it in :)

### DIFF
--- a/frontend/src/__tests__/UnorderedList.test.js
+++ b/frontend/src/__tests__/UnorderedList.test.js
@@ -2,12 +2,7 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import { ThemeProvider } from 'emotion-theming'
 import theme from '../theme'
-import {
-  UnorderedList,
-  Ul,
-  UlNone,
-  UlSquare,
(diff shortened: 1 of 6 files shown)


frontend/src/__tests__/UnorderedList.test.js
```diff
@@ -33,17 +28,5 @@ describe('Unordered List', () => {
       </ThemeProvider>,
       div,
     )
-    ReactDOM.render(
```
sastels: Might be worth testing one with a `listStyleType` passed in

evadb: 👍

PR #127, opened 2019-02-14
poke compliance api

### LINKED ISSUE #125: Trigger a compliance API check when you app deploys
The intent is to trigger a compliance API check whenever you app has deployed and ready. Because this is a non-deterministic event, the requirement is for the app to decide when the best point in time for that is.

The compliance watcher application exposes an internal kubernetes endpoint (https://github.com/cds-snc/report-a-cybercrime/blob/master/compliance/watcher-deployment.yaml#L34), which can be triggered by a POST request.

In other words, you need to make a POST request to:

`compliance-watcher.symmorfosi.svc.cluster.local:3000/run_compliance`

to trigger a compliance run. I would recommend putting it in here or thereabouts:

https://github.com/cds-snc/report-a-cybercrime/blob/master/frontend/src/index.js#L22

sastels:
resolves #125

### DIFF
--- a/frontend/src/index.js
+++ b/frontend/src/index.js
@@ -24,7 +24,9 @@ const _port = port()
 fetch(
   'http://compliance-watcher.symmorfosi.svc.cluster.local:3000/run_compliance',
   { method: 'POST', body: '' },
-)
+).catch(err => {
+  Logger.error(`Error triggering compliance api: ${err}`)
+})
 
 server.listen(_port, error => {
   if (error) {
--- a/frontend/src/index.js
+++ b/frontend/src/index.js
@@ -1,6 +1,7 @@
 import app from './server'
 import http from 'http'
 import { Logger } from '@cdssnc/logdriver'
+import fetch from 'node-fetch'
 
 const server = http.createServer(app)
 
@@ -19,6 +20,11 @@ const port = () =>
   )
 
 const _port = port()
+
+fetch(
+  'http://compliance-watcher.symmorfosi.svc.cluster.local:3000/run_compliance',
+  { method: 'POST', body: '' },
+)
 
 server.listen(_port, error => {
   if (error) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,11 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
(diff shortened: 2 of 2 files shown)


sastels:
@maxneuvians is this what you had in mind?

maxneuvians:
Agreed - I am not happy about this, it was just the most simplest thing I could come up with. The idea is that it is a fire and forget hook, so if it fails locally, who cares (clearly too much elixir speaking). 

https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook look like a good alternative approach. Hopefully can

maxneuvians:
Another option is to throw the trigger logic in an init container, does not concern the app logic at all, but it would no ensure that the app is up when the checks are run

maxneuvians:
To overcome the problem of using an init container and the main container not being ready when the checks run, the watcher app could receive a delay param that says, wait X seconds before running the check, because the main container should be up by then. Again if after X seconds the thing is not up, you have a problem. There is this https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/ but the hooks are no `Container is running` hooks

maxneuvians:
I will take a look at this here: https://stackoverflow.com/questions/35192712/kubernetes-watch-pod-events-with-api, to see if I can tap into a `Running` event

maxneuvians:
Maybe add a try/catch around it until I have a chance to review the other implementations.

sastels:
locally, with the catch, I get the following sent to the logs (as expected)
```
{ cloudEventsVersion: '0.2',
  contentType: 'text/plain',
  data:
   'Error triggering compliance api: FetchError: request to http://compliance-watcher.symmorfosi.svc.cluster.local:3000/run_compliance failed, reason: getaddrinfo ENOTFOUND compliance-watcher.symmorfosi.svc.cluster.local compliance-watcher.symmorfosi.svc.cluster.local:3000',
  eventID: '3f6e8aa1-d5cf-4605-b540-ffa48d2a8cc4',
  eventTime: '2019-02-15T19:45:47.044Z',
  eventType: 'com.github.cds-snc.error',
  eventTypeVersion: '1.0',
  source: '/' }
```

PR #128, opened 2019-02-15
Breadcrumbs

sastels:
make a breadcrumb component

Example on the second form (How did that Happen?)

### DIFF
--- a/frontend/src/components/breadcrumb/presets.js
+++ b/frontend/src/components/breadcrumb/presets.js
@@ -36,7 +36,9 @@ export const Breadcrumb = props => {
       {...props}
     >
       {props.children.map((child, index) => (
-        <BreadcrumbLi key={index}>{child}</BreadcrumbLi>
+        <BreadcrumbLi key={index}>
+          {child}&nbsp;{index === props.children.length - 1 ? null : '/'}&nbsp;
+        </BreadcrumbLi>
       ))}
     </BreadcrumbList>
   )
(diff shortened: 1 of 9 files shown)


sastels:
[Create breadcrumb component](https://trello.com/c/CCBA2YD9/146-create-breadcrumb-component)

frontend/src/components/breadcrumb/presets.js
```diff
@@ -0,0 +1,55 @@
...
+export const Breadcrumb = props => {
+  var paddingValue
+
+  paddingValue = 0
+
+  return (
+    <BreadcrumbList
+      fontSize={[1, null, 2]}
+      lineHeight={[1, null, 2]}
+      pl={paddingValue}
```
evadb: is passing a paddingValue necessary? It looks like it never changes from 0?

frontend/src/components/breadcrumb/presets.js
```diff
@@ -0,0 +1,55 @@
...
+      mb={4}
+      css={css`
```
evadb: Since list-style-type: none is set in the component itself, I think this can be removed

evadb:
image

Mobile margin looks a bit big if the breadcrumbs end up on multiple lines

frontend/src/components/breadcrumb/presets.js
sastels: hrm, what the heck is `pl` anyway? I stole that code from `ordered-list` but what is `pl` there?

sastels:

PR #145, opened 2019-02-19
make button component

### LINKED ISSUE #130: Make button component

sastels:
resolves #130


sastels:
@JuliannaR @evadb Isn't an `em` equal to the current font size?

frontend/src/components/button/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+BaseButton.propTypes = {
+  ...fontSize.propTypes,
+  ...lineHeight.propTypes,
+  ...space.propTypes,
+  ...color.propTypes,
+  ...fontWeight.propTypes,
+}
+
+export { Button } from './presets'
```
sleepycat: The relationship between index.js and presets.js seems a little circular. index.js > BaseButton > presets.js > Button > index.js. Should we combine these into a single file?

frontend/src/WhatHappenedForm.js
```diff
@@ -3,7 +3,7 @@ import { Trans } from '@lingui/macro'
 import { i18nMark } from '@lingui/react'
 import { Form, Field } from 'react-final-form'
 import styled from '@emotion/styled'
-// import Button from '@govuk-react/button'
```
sleepycat: :dancer:

JuliannaR:
> Check out my comments and let me know if you agree!

Yes, I think so.

frontend/src/components/button/index.js
sastels: Maybe? @evadb , thoughts?

evadb: In some cases with only one preset I think it makes more sense to combine everything. If we end up building a bunch of different button presets for whatever reason I can't imagine at the moment, I like the idea of having a separate file for organization

frontend/src/components/button/index.js
```diff
@@ -0,0 +1,27 @@
+import styled from '@emotion/styled'
+import { fontSize, lineHeight, space, color, fontWeight } from 'styled-system'
+import tag from 'clean-tag'
+
+export const BaseButton = styled(tag.button)`
+  font-family: ${({ theme }) => theme.fontSans};
+  margin: 0;
+  border: 0;
+  &:focus {
```
evadb: A thought: it might be an idea to pass a focus color as a prop so that the focus color can be customized if needed, with this color as the default prop?

frontend/src/WhatHappenedForm.js
evadb: 🎉

evadb:
@JuliannaR @sastels the answer to pt vs em is very boring in that I took what gov.uk was using :) Is a thing we can discuss down the road and it's easy to swap out

frontend/src/components/button/index.js
sastels: So I'm wondering if we want this vs having the focus color set in the theme and having all our components get it from there?

sastels:

frontend/src/components/button/index.js
sleepycat: Getting the focus colour from the theme seems to be more in line with the general model IMO.

sastels: I was thinking of not adding that here, but rather in a separate PR (and there we'd also use it for other relevant components)

PR #155, opened 2019-02-21
Breadcrumb added and test created

keithiancolbourne:


frontend/src/Screen3.js
```diff
@@ -102,18 +104,20 @@ const MyForm = () => (
...
       <Link to={'/'}>
         <Trans>Landing Page</Trans>
       </Link>
       <Link to={'/form1'}>
-        <Trans>What happened?</Trans>
+        <Trans>What Happened</Trans>
       </Link>
-      <Link to={'/form2'}>
-        <Trans>What was involved?</Trans>
+      <Link to="{'/form2'}">
```
sastels: those `"` are messing up the link. TBH, we probably don't need the curly braces, ie `<Link to="/form2">` works I think.

sastels:
Could you also add the breadcrumb to `Screen1`?

keithiancolbourne:
> Could you also add the breadcrumb to `Screen1`?

Yup, I am working on that but there's the issue of it not working with one child so trying to find a way to recode for that case. Maybe using React.children.map rather than props.children.map but I don't know enough about React yet to say. I need to take some tutorials before I can get too deep into it because I don't know what half the code is doing. Ha.

sastels:
Oh wow, yeah looks like we really don’t want to just use map on props.children
https://mxstbr.blog/2017/02/react-children-deepdive/

sastels:

sastels:
[Refactor existing discovery tool to use new components](https://trello.com/c/VZG90Fjw/115-refactor-existing-discovery-tool-to-use-new-components)

frontend/src/components/breadcrumb/presets.js
```diff
@@ -27,10 +27,10 @@ export const Breadcrumb = props => {
       mb={4}
       {...props}
     >
-      {props.children.map((child, index) => (
-        <BreadcrumbLi key={index}>
+      {React.Children.map(props.children, (child, index) => (
```
sastels: cool!

frontend/src/Screen1.js
```diff
@@ -32,16 +32,14 @@ const ListItem = styled('li')`
...
       <Link to={'/'}>
         <Trans>Landing Page</Trans>
       </Link>
-    </Breadcrumb> */}
-    <div>
-      <Text display="inline-block">Placeholder&nbsp;</Text>
-      <Text display="inline-block">for</Text>
-      <Text fontSize={[4, 4, 5]}>Breadcrumb</Text>
-    </div>
+      <Link to="/form1">
```
sastels: this should be a `<Text display="inline-block">` instead of the `Link`

PR #158, opened 2019-02-22
Add Top Banner component

evadb:
_finally_

This PR adds a responsive Top Banner component using the Container, GoCSig & LanguageSwitcher components. Discussed the issues of aria-label/aria-describedby/alt text etc with @JuliannaR, there's some kind of issue there with the screen reader not picking up the SVG as an image, but that's something we an solve down the line when we swap out for the RCMP logo and potentially add a link back to the RCMP website, changing the behaviour.

### DIFF
--- a/frontend/src/components/topbanner/index.js
+++ b/frontend/src/components/topbanner/index.js
@@ -1,0 +1,48 @@
+import PropTypes from 'prop-types'
+/** @jsx jsx **/
+import { jsx, css } from '@emotion/core'
+import { Container } from '../container'
+import GoCSignature from '../gocsig'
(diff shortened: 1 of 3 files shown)


frontend/src/components/topbanner/index.js
```diff
@@ -0,0 +1,48 @@
...
+      >
+        <LanguageSwitcher />
+      </Container>
+    </Container>
+  )
+}
+
+TopBanner.propTypes = {
+  lang: PropTypes.string,
+  bg: PropTypes.string,
```
sastels: should these be `required` or have defaults?

evadb: fixed :)

PR #161, opened 2019-02-28
Add footer

evadb:
This PR replaces the current footer with a mobile responsive footer components made with flexbox 🎉 
In order to do this, I needed to add the text alignment prop to the container and the display prop to the list items. I created a separate link preset for the footer links, and added them as children to the footer component, because it felt... wrong to bake them right into the footer.


sastels:
Getting a white bar under the footer on small screens (not big screens though)
image

frontend/src/components/unordered-list/presets.js
```diff
@@ -1,7 +1,7 @@
-import React from 'react'
 import PropTypes from 'prop-types'
 import { UnorderedList } from '.'
-import { css } from '@emotion/core'
+/** @jsx jsx **/
```
sastels: Does this do something? 🤔

frontend/src/theme/index.js
```diff
@@ -63,6 +63,13 @@ const theme = {
         color: '#2b8cc4',
       },
     },
+
+    footerLink: {
+      color: '#FFF',
```
sastels: Can we use something like `theme.colors.white` here?

sastels:
resolves #131

frontend/src/components/unordered-list/presets.js
evadb: It does! I made a mistake in the first place by not bringing in the JSX, otherwise the CSS prop isn't applied properly and the class shows up as [Object object], which... isn't ideal!

frontend/src/theme/index.js
evadb: good call! i'll open an issue to refactor WordMark & GoCSig to accept styled-system props

frontend/src/components/unordered-list/presets.js
sastels: oh yeah, I forgot we need to do that for emotion 10.

evadb: I did too! :joy:

PR #163, opened 2019-03-01
Refactor Header tests

evadb:
This PR adds a better test suite to the Header component. I've also changed the 'is' prop to be required by the Header component. This is because this prop sets the tag the Header will render as (h1, h2, h3 etc) and the Header component should never be used without it.


frontend/src/components/header/__tests__/Header.test.js
```diff
@@ -1,59 +1,115 @@
...
+  })
+
+  it("The 'is' prop sets correct header tag", () => {
+    let wrapper = mount(<Header is="h1" />).find('h1')
+    expect(wrapper.is('h1')).toBeTruthy()
+  })
+
+  it('H1 preset renders as h1', () => {
+    let wrapper = mount(<H1 />).find('h1')
+    expect(wrapper.is('h1')).toBeTruthy()
```
sastels: this is nice!

frontend/src/components/header/__tests__/Header.test.js
```diff
@@ -1,59 +1,115 @@
...
+        foo
+      </Header>,
     )
+
+    const test = getByText('foo')
+    expect(test).toHaveStyleRule('font-weight', 'bold')
+  })
+
+  it("The 'is' prop sets correct header tag", () => {
+    let wrapper = mount(<Header is="h1" />).find('h1')
```
sastels: Do you need all these `find`s here and in subsequent tests? Isn't wrapper just going to be the entire thing anyway?

evadb: If `shallow` rendering yes. Unfortunately I can't seem to use Shallow rendering for these? :( So I used `mount` which wraps it in a bunch of divs. So I used the finds to pick up the header

frontend/src/components/header/__tests__/Header.test.js
evadb: shamelessly stolen from @sleepycat

frontend/src/components/header/__tests__/Header.test.js
sastels: wait a sec, why are you using `enzyme` here, can't something from `react-testing-library` do the job?

evadb: I spent a while reading docs and looking things up, but was unable to find a matcher to check for HTML tag type. Enzyme had one so I decided to just use that

PR #183, opened 2019-03-11
Add initial test landing page

sastels:
New landing page, at url `/nlp`


frontend/src/NewLandingPage.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import styled from '@emotion/styled'
+import { Trans } from '@lingui/macro'
+import { H1 } from './components/header'
+import { P } from './components/paragraph'
+import { TrackPageViews } from './TrackPageViews'
+
+const CenterContent = styled('div')`
```
evadb: You could use the Container component here :O

sastels: Yes! Good call!

PR #184, opened 2019-03-11
Update most frontend dependencies.

sleepycat:
This commit omits Apollo related things which will follow with some associated
changes in another commit.

Minor change to a test file since Jest now warns about using async with a
describe block.


sastels:
This is helpful:
New styled system
- Adds long-form props for margin and padding (e.g. marginTop). Shorthand aliases still work as before.

keithiancolbourne:
Just curious...do you decide to update these deps based on a security vulnerability scan or what triggers an update like this?

sastels:
Not sure what tickled Mike this time, but on the VAC project we got in the habit of someone doing it every sprint. Basically, update everything that’s easy, and if there are updates that would break the app, make an issue for them instead.

sleepycat:
Big across-the-board updates like this are usually a "it's been a while" moment. Security driven updates are usually smaller and more focused.

Issue #139, opened 2019-02-19
Make label component for forms


sastels:

evadb:
addressed by https://github.com/cds-snc/report-a-cybercrime/pull/191

PR #191, opened 2019-03-14
Add label

evadb:
What it says on the tin! Very similar to Input, this adds a Label component hooked up with styled-system props and uh, not much else!

### DIFF
--- a/frontend/src/components/label/index.js
+++ b/frontend/src/components/label/index.js
@@ -1,0 +1,37 @@
+import styled from '@emotion/styled'
+import tag from 'clean-tag'
+import {
+  fontSize,
+  fontWeight,
+  lineHeight,
+  space,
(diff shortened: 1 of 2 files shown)


sastels:
resolves #139

frontend/src/components/label/index.js
```diff
@@ -0,0 +1,39 @@
...
+  ${borderColor};
+  ${width};
+  ${height};
+  ${position};
+  ${zIndex};
+  ${top};
+  ${right};
+  ${bottom};
+  ${left};
+`
```
sastels: What about adding some defaults?

evadb: I'm honestly not sure what our defaults would look like right now! I figure we can add them as use cases come up?

PR #192, opened 2019-03-15
Added jsx-a11y Linter 

khalidelaggan:
This commit/linter is just for testing and exploration purposes to see what can work as a standard Accessibility linter tool.

Notes:
- It will be only tested with all A11y baked in rules within the linter for common aria* validations
- Any custom rules for styled components will be added later with more feedback from the team :)


sastels:
Looks like most of the errors were issues in the tests - we probably want to add the test dirs to an ignore file. then two other errors in the app that are probably legit.

PR #195, opened 2019-03-15
Component Swapping 

keithiancolbourne:
Swapping out locally created components and native elements for those from our component library.


frontend/src/Screen1.js
```diff
@@ -46,21 +33,30 @@ export const Screen1 = () => (
...
+      <Li>
         <Trans>How were you affected?</Trans>
-      </ListItem>
-    </ul>
+      </Li>
+    </UnorderedList>
 
-    <Warning>
+    <Container
+      css={css`
```
evadb: I need to add some better documentation to these, sorry! But you should be able to do all of this block of css with https://github.com/styled-system/styled-system props! You can check out the link, or the Top Banner/Footer components to see how to use props to set there :)

frontend/src/Screen2.js
```diff
@@ -67,49 +44,56 @@ const MyForm = options => (
...
+            </H4>
+            <Container>
               <Field name="whatWasInvolvedOther">
                 {({ input, meta }) => (
-                  <div>
+                  <Container>
                     <TextArea {...input} placeholder="" />
-                    <ErrorMessage>
+                    <Container
+                      css={css`
```
evadb: Same comment as above!

frontend/src/Screen2.js
```diff
@@ -128,7 +112,12 @@ const options = {
 }
 
 export const Screen2 = () => (
-  <CenterContent>
+  <Container
```
evadb: Hmm, for these, it might be worth adding a preset to the Container, so we can just export and reuse CenterContainer. Doesn't have to be done for this PR, but it's an idea for a future PR :)

frontend/src/Screen1.js
```diff
@@ -46,21 +33,30 @@ export const Screen1 = () => (
     </H1>
 
     <TrackPageViews />
-    <ul>
-      <ListItem>
+    <UnorderedList>
```
evadb: I'd use the Ul preset here!

evadb:
This is awesome Keith! Just a couple of small comments 👍

keithiancolbourne:
Thanks Eva! I will get back to these after I'm done what I'm working on now

frontend/src/Screen2.js
```diff
@@ -85,32 +62,36 @@ const MyForm = options => (
...
+                    </Label>
                   )
                 })}
-              </div>
-            </div>
-            <LabelFormat>
-              <label htmlFor="whatWasInvolvedOther">
+              </Container>
+            </Container>
+            <H4>
```
evadb: Small comment! Headers should be used for page structure only. In this case, since you're after larger font, you would want to use a paragraph with the styles applied to make the text appear how you want

keithiancolbourne: For sure, makes sense. I was just swapping out what was there but should have been thinking of this. Thanks!

PR #197, opened 2019-03-18
don't cache graphql calls in service-worker

sastels:
the graphql POSTs were causing the `service-worker` code to throw an error. In this PR we restrict the caching to GETs.


evadb:
This looks fine to me but nominating @sleepycat for the actual reviewing because he knows way more about this

Issue #205, opened 2019-03-22
Scroll bars on landing page


sleepycat:
The landing page is longer than the browser window resulting in scroll bars.
The culprit is `min-height: calc(100vh - 55px);` in src/App.js, but finding an approach to sticky footers that doesn't require fixed pixel heights might be worth exploring.
![offscreen](https://user-images.githubusercontent.com/109692/54853798-3ac98900-4cc7-11e9-8c01-f4f36eed1242.png)

sastels:
even more disturbing: no bar on Firefox, has the bar on both Safari and Chrome.

evadb:
seems fun - I'm going to work on this

evadb:
it fixed

PR #207, opened 2019-03-25
Fix sticky footer

evadb:
This PR addresses https://github.com/cds-snc/report-a-cybercrime/issues/205 by replacing the height minus footer height approach to sticky footer with a flexbox based approach. I also did some clean up work while I was in there just to make stuff easier to debug by wrapping the header/main/footer in their proper section tags and replacing one of the root divs with a fragment. Basically I got bored of digging into a million divs to find what I wanted.


frontend/src/App.js
```diff
@@ -11,15 +11,12 @@ import { GET_LANGUAGE_QUERY } from './utils/queriesAndMutations'
...
-const Content = styled('div')`
-  min-height: calc(100vh - 55px);
-`
+import { Content } from './components/container'
 
 const App = () => (
   <Query query={GET_LANGUAGE_QUERY}>
     {({ data: { language } }) => (
-      <div>
+      <React.Fragment>
```
sastels: 👍

Issue #204, opened 2019-03-22
[react-ga] gaTrackingID is required in initialize()


sleepycat:
It appears that the Google Analytics ID is not being passed into the app.

sastels:
That's odd. Was it ever? I can't seem to find in in our Google Analytics account...

sastels:
huh, looks we don't have secrets for the frontend code.

sastels:
after talking IRL it looks like the problem is that the ID isn't in the cloud build container when the frontend's docker image is built.

sastels:
@sleepycat fixed this in #350 🎉

PR #209, opened 2019-03-27
add current GA id as default

sastels:
addresses #204 

for now, add the current GA id as the default. note that the GA id is not confidential (it's included in the bundle)

### DIFF
--- a/frontend/src/TrackPageViews.js
+++ b/frontend/src/TrackPageViews.js
@@ -3,7 +3,7 @@ import ReactGA from 'react-ga'
 import Component from '@reach/component-component'
 import { Location } from '@reach/router'
 
-const GA = process.env.RAZZLE_GOOGLE_ANALYTICS_ID || ''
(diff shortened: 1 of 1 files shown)


sleepycat:
I think this should be tackled in the cloudbuild file. Maybe with [substitutions](https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values) or env vars.

sastels:
> I think this should be tackled in the cloudbuild file. Maybe with [substitutions](https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values) or env vars.

I looked into both of these, but since it’s the client that needs the id, not the server, I figured it’d be easiest to just add it to the code directly. Or is there an easy way with Razzle to have the server add variables to the bundle?

sastels:
> So I reconsidered, I think I'm letting the perfect become the enemy of the good here. Let's get this in there and get some stuff done and if we want to loop back on this later we can.

This was my thinking. 💯

PR #210, opened 2019-04-01
WIP: Create input for accepting identifiers

sleepycat:
This is some exploratory work on creating a single field that understands what kind of identifier it contains and "does the right thing".
As it stands, determining "the right thing" is relative to this components use within React Final Form, so next steps were going to be trying to use the two together to see if this API made any actual sense.


sleepycat:
Closing this since it was informational rather than something to be merged.

Issue #206, opened 2019-03-25
Mobile styling issues


evadb:
![Screen Shot 2019-03-25 at 5 51 16 PM](https://user-images.githubusercontent.com/1545806/54956580-af4e3300-4f26-11e9-9ccb-5ac36af6115b.png)

Noticing some issues in the mobile styling of pages, especially on pages with forms

sastels:
I think the width prop in the `Container`s is doing something wonky.

sastels:
fixed now

evadb:
email blast from the past

sastels:
word.

PR #215, opened 2019-04-03
New first page

samsadasivan:
S and S's first page design for hypothesis 1


keithiancolbourne:
`const CenterContent = styled('div')`
  max-width: 750px;
  margin: auto;
``

Do we want to use a container with css props instead of creating CenterContent here?

sastels:
> `const CenterContent = styled('div')`
> `max-width: 750px;`
> `margin: auto;`
> 
> Do we want to use a container with css props instead of creating CenterContent here?

Almost certainly yes. Except that containers are slightly messed up right now (#206) - I might take a stab at that one next.

evadb:
@sastels @keithiancolbourne I think my PR coming up will add a good preset for CenterContent using Container component instead

PR #224, opened 2019-04-05
Remove unneeded routes

sleepycat:
Page1 is now mounted at /
Summary is now at /summary
Page1 continue > /summary


sastels:
Probably a good PR to 🔥 the old stuff too (Screen1, etc). Maybe keep the `NewLandingPage` component around.

PR #228, opened 2019-04-08
tweak path names

anikbrazeau:


sastels:
You also have to change 
https://github.com/cds-snc/report-a-cybercrime/blob/path-names/frontend/src/Page1.js#L45
and
https://github.com/cds-snc/report-a-cybercrime/blob/path-names/frontend/src/Page2.js#L21

PR #231, opened 2019-04-08
Language change for precision

anikbrazeau:

PR #233, opened 2019-04-08
Fix code bedevilment

sastels:


lchski:
Anik covers this in #231.

PR #234, opened 2019-04-09
use theme font in language switcher button

sastels:
we just had "inherit" for the font, which was giving us the default font.

### DIFF
--- a/frontend/src/LanguageSwitcher.js
+++ b/frontend/src/LanguageSwitcher.js
@@ -2,21 +2,8 @@ import React from 'react'
 import { ApolloConsumer } from 'react-apollo'
 import styled from '@emotion/styled'
 import PropTypes from 'prop-types'
+import { ButtonLink } from './components/button-link'
 
-const ButtonLink = styled('button')`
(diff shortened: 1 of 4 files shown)


frontend/src/LanguageSwitcher.js
```diff
@@ -4,9 +4,10 @@ import styled from '@emotion/styled'
 import PropTypes from 'prop-types'
 
 const ButtonLink = styled('button')`
+  font-family: ${({ theme }) => theme.fontSans};
+  font-size: ${({ theme }) => theme.fontSizes[2]};
```
evadb: Being a bit nitpicky here - would it be better to pass in the fontSizes prop and then set it on the ButtonLink component instead of just passing fontSizes[2]? Would make it easier to change font sizes later (and I'm thinking we may just want to extract ButtonLink later)

sastels: yeah, probably. I was thinking that we'd want to make a proper styled-systems component at some point, but not before tomorrow. I'll pull out the font size prop as you suggest now though.

PR #244, opened 2019-04-10
Jest puppeteer integrations

khalidelaggan:
I will be more than happy to jump into a quick call to explain things if needed :)
Theres are few thing to be added later to ensure tests will run in the proper Env setup


sleepycat:
As discussed with @khalidelaggan jest-axe is going to be explored. Closing for now.

PR #245, opened 2019-04-10
Removed eslint-plugin-jsx-a11y linter

khalidelaggan:
Reasons for removal can be found here:
https://drive.google.com/open?id=16B203noMuj_6dC95tjzUH4_ViMT1yZPo9ijM2kJUW2I


sleepycat:
Hey @khalidelaggan I don't have any illusions that this linter is in anyway a comprehensive a11y solution, but it's admittedly small upside comes with 0 downside, so I think it makes sense to keep it and use it alongside other solutions.
Closing this for now. Happy to discuss further though!

PR #249, opened 2019-04-11
Replace clean-tag with custom util

evadb:
This PR replaces clean-tag and clean-element with a custom util cleanProps, and emotion's built in shouldForwardProp util. Basically, the cleanProp util compares the incoming prop to the list of styled system props, and returns true to shouldForwardProp if it is not in the list. A prop that is forwarded will appear in the rendered HTML. The goal of this util is to make sure no styled-system props appear in the final render, as they are only useful for generating the CSS.


evadb:
@sleepycat hence the WIP! :) Took a bit to get the tests right without suffering, but next step is to replace clean-tag everywhere with this and then remove clean tag

frontend/locale/fr/messages.json
```diff
@@ -27,4 +27,4 @@
   "a website": "d’un site web",
   "an email address": "une adresse courriel",
   "loading...": "chargement en cours..."
-}
\ No newline at end of file
+}
```
sastels: good catch

frontend/src/utils/cleanProps.js
```diff
@@ -0,0 +1,100 @@
+export const cleanProps = prop => {
+  if (!styledSystemProps.includes(prop) && prop !== 'as') return true
+  return false
+}
+
+export const styledSystemProps = [
```
sastels: We'll have to remember to update this when `styled-system` changes.

evadb: the good news is the actual list of props is very unlikely to change

PR #250, opened 2019-04-12
Label graph

sastels:
Add labels and title to graph, and use the theme font.

image


frontend/src/Stats.js
```diff
@@ -11,28 +14,52 @@ import {
...
+                  data={this.props.data}
+                  margin={{
+                    top: 0,
+                    right: 0,
+                    left: -20,
+                    bottom: 0,
+                  }}
+                >
+                  <CartesianGrid strokeDasharray="2 2" />
+                  <XAxis fontFamily={theme.fontSans} dataKey="date" />
```
sastels: Using the `ThemeProvider` didn't seem to work here, unless I was doing it the wrong way.

frontend/src/components/header/index.js
```diff
@@ -11,6 +18,7 @@ export const Header = styled(tag)`
   ${space};
   ${color};
   ${fontWeight};
+  ${textAlign};
```
sastels: wanted to be able to center a header

frontend/src/Stats.js
```diff
@@ -11,28 +14,52 @@ import {
...
+                  <YAxis
+                    dataKey="total"
+                    fontFamily={theme.fontSans}
+                    label={{
+                      value: i18n._(t`Reports per day`),
```
sastels: Couldn't just use `<Trans>` here, instead used the approach suggested in https://lingui.js.org/tutorials/react-patterns.html#element-attributes-and-string-only-translations

frontend/src/Stats.js
anikbrazeau: Can we make these changes in French too?

Title EN: Reports similar to yours
Title FR: Rapports semblables au vôtre

Y-axis EN: Reports per day
Y-axis FR: Rapports par jour

sastels: Right now the French (just from DeepL) is:
```
"Reports per day": "Rapports par jour",
"Reports similar to yours": "Des rapports similaires aux vôtres",
```
Can change the French title as you suggest 👍

sastels: image

keithiancolbourne:
Looks good! Are any tests needed?

sastels:
> Looks good! Are any tests needed?

I figured no. Just adding labels to the graph 🤷‍♂️

keithiancolbourne:
> > Looks good! Are any tests needed?
> 
> I figured no. Just adding labels to the graph 🤷‍♂️

Yeah, same. Only thing comes to mind is an accessibility test. i.e. checking to see if there's actually a label kinda thing

PR #252, opened 2019-04-12
spelling out RCMP acronym

anikbrazeau:


keithiancolbourne:
Should it be ...Police (RCMP) **needs** or **need** ?

sastels:
> Should it be ...Police (RCMP) **needs** or **need** ?

I think RCMP is singular, so “needs”. Unless I'm wrong. Very possible.

PR #262, opened 2019-04-18
Clean up cache creation

sleepycat:
The Apollo cache was being created with default values for the narrative
gathering tool. This commit removes those and defaults the cached value for
language to "fr" if it can find it in the navigator.languages array.

### DIFF
--- a/frontend/src/client.js
+++ b/frontend/src/client.js
@@ -21,11 +21,7 @@ const cache = new InMemoryCache()
 // Add defaults for form fields
 cache.writeData({
   data: {
-    language: 'en',
-    whatHappened: '',
-    whatWasInvolved: [],
-    whatWasInvolvedOther: '',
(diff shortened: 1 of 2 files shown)


frontend/src/client.js
```diff
@@ -21,11 +21,7 @@ const cache = new InMemoryCache()
 // Add defaults for form fields
 cache.writeData({
   data: {
-    language: 'en',
-    whatHappened: '',
-    whatWasInvolved: [],
-    whatWasInvolvedOther: '',
-    howWereYouAffected: '',
+    language: navigator.languages.includes('fr') ? 'fr' : 'en',
```
sastels: 🎉

sastels:
I set the language of Chrome to French - the app loads in French, but after a second snaps to English.

PR #265, opened 2019-04-18
use demo data for offline graph

sastels:
Will make it easier to test with users - won't need internet 🤞 

Also, the chart page will display properly on `npm run dev` / Heroku

### DIFF
--- a/frontend/src/Stats.js
+++ b/frontend/src/Stats.js
@@ -3,7 +3,6 @@ import { Query } from 'react-apollo'
 import PropTypes from 'prop-types'
 import { I18n } from '@lingui/react'
 import { t, Trans } from '@lingui/macro'
-import { P } from './components/paragraph'
 import { H3 } from './components/header'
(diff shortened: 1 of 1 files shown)


frontend/src/Stats.js
```diff
@@ -73,18 +76,11 @@ export const Stats = () => (
...
-            Error connecting to database:
-            <br />
-            {error.message}
-          </P>
-        )
-
-      let { stats } = data
-      return <Chart data={stats.identifierFlaggingsWithin.summary} />
+      let summary = undefined
+      if (error) summary = demoData
```
evadb: Nitpicky - should we specify the error being the error we get if not connected to the internet? Or do we want just any error leading to demo data right now?

sastels: Huh, yeah, I like this idea 💯

sastels:
After talking IRL I think we probably don't want to do this (in particular, as we use the backend more and more, following this path would require constant "offline" updates.

Issue #255, opened 2019-04-16
Return better data to feed summary graph


sleepycat:
Currently the identifier visualization used in our prototype is powered by the following data:
```json
{
  "data": {
    "stats": {
      "identifierFlaggingsWithin": {
        "identifier": "555-555-5555",
        "summary": [
          {
            "date": "2019-04-01",
            "total": 1
          },
          {
            "date": "2019-04-02",
            "total": 2
          },
          {
            "date": "2019-04-03",
            "total": 2
          },
          {
            "date": "2019-04-04",
            "total": 1
          },
          {
            "date": "2019-04-05",
            "total": 5
          },
          {
            "date": "2019-04-09",
            "total": 1
          },
          {
            "date": "2019-04-15",
            "total": 1
          }
        ]
      }
    }
  }
}
```
This data is a naively generated list of days we have data for, rather than a report count for dates within some time period. The result is a misleading chart suggesting continuity that would not actually exist if all the dates were represented.

![image](https://user-images.githubusercontent.com/109692/56231647-fd99b200-604c-11e9-92a2-8a5ca61f3bc0.png)

Suggested fix: modifying `identifierFlaggingsWithin` to accept start/end dates along with an identifier (it's name suggests that this was the intent of that function anyway):
```graphql
query {
  stats {
    identifierFlaggingsWithin(startDate: "2019-04-02", endDate: "2019-05-02" identifier: "555-555-5555") {
      identifier
      summary {
        date
        total
      }
    }
  }
}
```
We may want to limit the date range to something reasonable, but if we do that we might be better off making an explicit query for it:
```graphql
query {
  stats {
    flaggingsWithinTheLastThreeMonths(identifier: "555-555-5555") {
      identifier
      summary {
        date
        total
      }
    }
  }
}
```

PR #267, opened 2019-04-24
Serve better summary data

sleepycat:
This commit adds a new function to the graphql api that returns proper data for
the summary graph. "Proper" in this case meaning that it the user supplies a
date range and the summary includes days with 0 reports within the range.

This addresses #255 and will be followed with a commit to get the frontend
using this new function.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -22,7 +22,11 @@
     "dotenv-safe": "^6.1.0",
     "express": "^4.16.4",
     "express-graphql": "^0.7.1",
-    "graphql": "^14.2.1"
+    "graphql": "^14.2.1",
+    "graphql-custom-types": "^1.5.0",
+    "lodash.unionby": "^4.8.0",
+    "moment": "^2.24.0",
+    "moment-range": "^4.0.2"
   },
   "devDependencies": {
     "eslint": "^5.16.0",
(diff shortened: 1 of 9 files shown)


api/package.json
```diff
@@ -22,7 +22,11 @@
     "dotenv-safe": "^6.1.0",
     "express": "^4.16.4",
     "express-graphql": "^0.7.1",
-    "graphql": "^14.2.1"
+    "graphql": "^14.2.1",
+    "graphql-custom-types": "^1.5.0",
+    "lodash.unionby": "^4.8.0",
+    "moment": "^2.24.0",
```
sastels: sad panda, this is a big module.

api/src/dbinit.js
```diff
@@ -17,6 +23,24 @@ const dbinit = async db => {
...
+            WITH COUNT INTO count
+            RETURN {
+              date: day,
+              total: count
+            }
+      `
+      let results = await db.query(query)
+      let summaries = await results.all()
+      let dates = generateDateObjects(startDate, endDate, { total: 0 })
+      return sortByDateAttribute(uniqueArray(unionBy(summaries, dates, 'date')))
```
sastels: lodash FTW! nice.

PR #269, opened 2019-04-24
Remove unused global vars

sleepycat:
This stops eslint from yelling


sastels:
:/ Odd, why isn't cloudbuild here? 
image

PR #272, opened 2019-04-25
Fix apostrophe on summary page

lchski:
From `‘` to `’` (correct orientation)


lchski:
Will need to update the translation strings, too.

PR #300, opened 2019-05-08
Switching the order of these cards

anikbrazeau:
No link consequence


sastels:
don't we have to switch the images too?

PR #305, opened 2019-05-09
Links stay blue

sastels:
As requested by @danprime , cards to visited pages don't get purple text.


evadb:
@sastels GO TO BED

sastels:
☝️ that's not a review!

PR #321, opened 2019-06-04
Upgrade api packages

sastels:
can still get data in and out via frontend app


sastels:
> Looks straightforward enough. Most are minor releases but a couple are major (eslint-plugin-node and husky)? No concerns there?

Everything still worked (using docker-compose) after the upgrades.

PR #327, opened 2019-06-13
Move Kubernetes config into a separate folder

sleepycat:
This commit adds the platform folder which now holds all config. This config is
set up to use kustomize so that we can generate variations on it for various
environments we use.


sastels:
Is this right?
image

sleepycat:
Ooop, good catch! I'll remove that.

PR #332, opened 2019-06-17
use new react testing library

sastels:
#331 didn't include a few tests that had been added after #331 branched from master.


sastels:
tests ran
image

PR #362, opened 2019-06-21
Replace kaniko with docker build

sleepycat:
Kaniko builds a bit faster and simplifies cloudbuild config a little by
automatically pushing images. Unfortunately it seems to build images that have
missing files, which then causes breakage downstream.
As an example:
https://github.com/GoogleContainerTools/kaniko/issues/362

### DIFF
--- a/api/cloudbuild.yaml
+++ b/api/cloudbuild.yaml
@@ -1,7 +1,7 @@
 steps:
-  - name: 'mikewilliamson/usesthis-ci'
+  - name: mikewilliamson/usesthis-ci
     id: start_arangodb
-    entrypoint: '/bin/sh'
+    entrypoint: /bin/sh
     args: [
         '-c',
         "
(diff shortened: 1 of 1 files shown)


api/cloudbuild.yaml
```diff
@@ -37,12 +37,20 @@ steps:
       - 'MINIO_SECRET_KEY=$_MINIO_SECRET_KEY'
       - 'MINIO_BUCKET_NAME=$_MINIO_BUCKET_NAME'
 
-  - name: gcr.io/kaniko-project/executor:debug
+  - name: 'gcr.io/cloud-builders/docker'
```
sastels: need the quotes? or is that just Prettier?

sleepycat: There isn't really a good reason for that. I'll get rid of it.

sleepycat:
Yeah I ran into similar problems there too and changed it before.

PR #368, opened 2019-06-24
Testing

DianeLiu2019:
add "cancel report" button and french version for the ifvictim.js and suspectinfo.js file


CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@ If this is your first time contributing on GitHub, don't worry! Let us know if y
 
 **Do not post any security issues on the public repository!** See [SECURITY.md](SECURITY.md)
 
-______________________
+---
```
sastels: We should figure out why husky is changing files that we aren't touching in our PR :(

PR #370, opened 2019-06-25
[WIP] add a11y scanning

sastels:
- add the a11y-scanning GitHub action
- fix the problems it flagged

The scanner sometimes has issues and needs to be restarted, but I think it's a worthwhile addition to our testing. Let's try it for a bit and see if it's useful.


sastels:
I might get back to this later, but closing for now...

PR #435, opened 2019-07-03
move victim line

keithiancolbourne:


anikbrazeau:
Seems like this duplicates the question on the page? Not sure if this is because of other changes we've made, and a conflict?

Latest copy should be: 
= "Are you reporting a scam that happened to you?"
= "Êtes-vous la personne visée par la fraude que vous signalez?"

keithiancolbourne:
Assuming this was closed by accident @DianeLiu2019 ? Reopening.

keithiancolbourne:
> Seems like this duplicates the question on the page? Not sure if this is because of other changes we've made, and a conflict?
> 
> Latest copy should be:
> = "Are you reporting a scam that happened to you?"
> = "Êtes-vous la personne visée par la fraude que vous signalez?"

@anikbrazeau This PR was just to move the line to a different place, I'm not sure I fully understand the comment -- in any case I'll talk to you in person about it and if we need to change or remove txt we can do in another PR. I'll merge this for now. Thanks!

PR #442, opened 2019-07-05
make label bold on confirmation page

DianeLiu2019:
make label bold on confirmation page


sastels:
Looks good! Can you update the lingui file (`locale/fr/messages.js`) as well?

PR #443, opened 2019-07-10
Save data

### LINKED ISSUE #379: send data from frontend to api

sastels:
send data from frontend to api (and thus to the database).

scramble the PII fields before sending to api.

resolves #379


sastels:
> We don't want to use our custom types for phone, email?

Almost certainly, though if we're going to do that we probably want some validation on the front end. Otherwise the user would type stuff in, submit, and it could just get dropped by the system without the user getting a chance to fix (or even knowing it was dropped). I figured this would deserve its own PR.

PR #482, opened 2019-07-17
Encrypt secrets per environment

sleepycat:
This commit creates a new set of secrets for Minikube that we can use for
running the app locally. The existing secrets have all been moved into the AKS
overlay.


sastels:
verified that works on minikube works using the minikube private key we have in LastPass.

still having issues getting it working on minikube from scratch (ie creating my own keypair and encrypting secrets myself. I think it's just that I don't have a `credentials.json` file.

PR #489, opened 2019-07-23
delete old font " roboto-regular-webfont.woff" and

DianeLiu2019:
delete old file"roboto-regular-webfont.woff" and "roboto-regular-webfont.woff2"


script.js
```diff
@@ -0,0 +1,7 @@
+import http from "k6/http";
```
sastels: :) I don't think you meant to include this file.

DianeLiu2019: tks

PR #498, opened 2019-07-30
remove links in stepper bar. make the stepper numbers are not clickable

DianeLiu2019:
remove links in stepper bar. make the stepper numbers are not clickable


sastels:
Looks good! Just missed the confirmation page.

PR #508, opened 2019-07-31
fix the problem:choosing a date on the "Describe the scam" page reset…

DianeLiu2019:
fix the problem:choosing a date on the "Describe the scam" page reset the other fields in the form


frontend/src/forms/ScamInfoForm.js
```diff
@@ -60,7 +59,7 @@ export class ScamInfoForm extends Component {
       <ApolloConsumer>
         {client => (
           <Form
-            initialValues={getScamInfo(client)}
+            // initialValues={getScamInfo(client)}
```
sastels: Hrm. We added this so that you can go back to this form from the Confirmation page. Without this line, when you come back the fields are all empty. This fixes the date picker issue, though, which seems to mean that maybe the form is getting rerendered (and these values stuffed in) when you pick a date? Maybe when we pick a date we have to save the form values.

huh, looking at the code, when we set the date we call `handleChange`, which sets the state, which probably causes the rerender.

sastels: You could make a `saveData` function in `ScamInfoPage` similar to `submitAndNavigate` and use that in the `ScamInfoForm` to save the data in `handleChange`. Then when the page got rerendered it would pull the updated data rather than the old stuff.

Issue #521, opened 2019-08-07
Move prototype 1 to p1


sastels:
- [x] Move current pages to p1 directories
- [x] Access via /p1/ paths
- [x] add a "source" field to the api that gets set to "p1" or "p2"

PR #522, opened 2019-08-08
Move p1

sastels:
re #521 

Move prototype 1 pages and forms to src/p1/... and urls to /p1/...

https://rac-prototype-pr-522.herokuapp.com/p1


sastels:
not sure why the k8s review app action decided to run? No config code here for it.

PR #523, opened 2019-08-08
azure pipeline

keithiancolbourne:
First stab at PR for azure pipeline integration


keithiancolbourne:
> LGTM? Should we also 🔥 the cloudbuild.yaml files in frontend and api?

Yeah I think we should. I kept it in for now until we see that azure works.

PR #540, opened 2019-08-09
Pull from gcr.io

sleepycat:
Switching back to grc.io

### DIFF
--- a/platform/base/kustomization.yaml
+++ b/platform/base/kustomization.yaml
@@ -22,7 +22,7 @@ resources:
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
-- name: acrrcmpcybercrime.azurecr.io/api
-  newTag: master-c110a7b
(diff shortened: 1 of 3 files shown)


platform/base/kustomization.yaml
```diff
@@ -22,7 +22,7 @@ resources:
...
 kind: Kustomization
 images:
-- name: acrrcmpcybercrime.azurecr.io/api
-  newTag: master-c110a7b
-- name: acrrcmpcybercrime.azurecr.io/frontend
-  newTag: master-cb3a090
+- name: gcr.io/report-a-cybercrime-alpha/api
+  newTag: master-f2ceb01
+- name: gcr.io/report-a-cybercrime-alpha/frontend
+  newTag: master-f2ceb01
```
sastels: these tags shouldn't be the same (or does it not matter)?

sleepycat: Looks like they were both built from the same commit (thats the branch-shortSHA). The build triggers need some tweaking, but that's definitely possible.

PR #543, opened 2019-08-09
Get GKE working

sleepycat:
This commit adds the secrets needed to get the app up and running on GKE. There
is also some documentation on the process in the README.


platform/overlays/gke-no-tls/arango-secrets.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: bitnami.com/v1alpha1
```
sastels: Aren't we generating these files each time we make the new key?

sleepycat:
Merging this in for now, but based on conversations with @sastels IRL, we'll revisit soon to make this a little more approachable.

PR #546, opened 2019-08-12
Update README.md with front-end instructions

danprime:
Add instructions/command to run the front-end in dev mode.


danprime:
I had forgotten how to to run the server for the frontend only. I couldn't find it in the README.

Issue #530, opened 2019-08-08
P2 contact info page


sastels:
<img width="109" alt="image" src="https://user-images.githubusercontent.com/8228248/62723020-b493b200-b9dd-11e9-99cf-fe5be2447060.png">

danprime:
![](https://github.trello.services/images/mini-trello-icon.png) [Build Pages for Prototype 2](https://trello.com/c/pGspCQOM/433-build-pages-for-prototype-2)

PR #549, opened 2019-08-12
Contact Info Page for P2

danprime:
Committer: Daniel Tse <daniel.tse@cds-snc.ca>


danprime:
Will need to review it under /p2/contactinfo

danprime:
Linked to Issue #530

frontend/src/Home.js
```diff
@@ -45,7 +45,7 @@ export const Home = () => (
         <FilesToUpload path="/p1/uploadfilesquestion" />
         <FileUploadPage path="/p1/uploadfiles" />
         <Willbecontacted path="/p1/contactinfoquestion" />
-        <ContactInfoPage path="/p1/contactinfo" />
+        <ContactInfoPage path="/p2/contactinfo" />
```
sastels: This will get rid of Prototype 1's contactinfo page - instead we want to do something like
```
 import { ContactInfoPage as P2ContactInfoPage } from './p2/ContactInfoPage'
```
above (leaving the p1 import alone) and here add
```
<P2ContactInfoPage path="/p2/contactinfo" />
```
while leaving `/p1/contactinfo` in place.

sastels: This will get rid of Prototype 1's contactinfo page - instead we want to do something like
```
 import { ContactInfoPage as P2ContactInfoPage } from './p2/ContactInfoPage'
```
above (leaving the p1 import alone) and here add
```
<P2ContactInfoPage path="/p2/contactinfo" />
```
while leaving `/p1/contactinfo` in place.

danprime: Fixed/renamed in the next push.

frontend/src/Home.js
```diff
@@ -45,7 +45,7 @@ export const Home = () => (
         <FilesToUpload path="/p1/uploadfilesquestion" />
         <FileUploadPage path="/p1/uploadfiles" />
         <Willbecontacted path="/p1/contactinfoquestion" />
-        <ContactInfoPage path="/p1/contactinfo" />
+        <P2ContactInfoPage path="/p2/contactinfo" />
```
sastels: ah, we need to leave in the p1 page as well :)

danprime: This makes sense

PR #559, opened 2019-08-15
Updating landing page content

anikbrazeau:
Hope this will set clearer expectations about the service is/is not and what be asked in the pages that follow.


anikbrazeau:
[Content for P2](https://trello.com/c/g0wY9kBi/426-content-for-p2)

PR #561, opened 2019-08-15
Integrate more pages with confirmation page

sastels:
https://rac-prototype-pr-561.herokuapp.com/p2/confirmation

* hook up the data from new pages
* make pages get data from cache (so can navigate from confirmation page back)


sastels:
let me pull in the contact info stuff too...

PR #565, opened 2019-08-16
P1 glitch -- French line disappeared

anikbrazeau:
Adding text back in to replace 2222


anikbrazeau:
[Content for P2](https://trello.com/c/g0wY9kBi/426-content-for-p2)

PR #568, opened 2019-08-16
add a mutation for prototype 2

sastels:
a mutation for the frontend to call to save Prototype 2 data into the database.


danprime:
This looks good. Flagging that the files and file descriptions are both arrays of strings ([String]). This should work for the prototype but will definitely need some time to look at it again.

sastels:
> This looks good. Flagging that the files and file descriptions are both arrays of strings ([String]). This should work for the prototype but will definitely need some time to look at it again.

Right now the files array is an array of file names. We’re not actually saving the files, when we do we’ll probably have them in Azure blob storage with some sort of file descriptor saved in arango. Probably want to refactor it into an array of objects, where each object has a filename, some sort of file location descriptor, and the description the user entered.

PR #569, opened 2019-08-16
Handle CAFC form data better

sleepycat:
This commit gets the server mimicing more closely the behaviour of the original server.
This basically translates to retaining a report state in a global variable
which is modified by a succession of requests, and finally reset when the form
is finally submitted.


danprime:
The context for this is the CAFC clone and server meant the CAFC server.

baseline/server.js
```diff
@@ -1,268 +1,297 @@
...
+    businessOrganizationType: null,
+    emails: [{ address: null }],
+    g1: null,
+    g2: null,
+    languagePreferred: null,
+    languagesSpoken: [],
+    otherName: null,
+    phoneNumbers: [
+      { cleanNumber: null, extension: null, number: null, phoneType: null },
+      { cleanNumber: null, extension: null, number: null, phoneType: null },
```
danprime: Why the extra/duplicate object here?

baseline/server.js
```diff
@@ -1,268 +1,297 @@
...
+    ageRange: null,
+    dob: null,
+    gender: null,
+    isAnonymous: null,
+    isBusiness: null,
+  },
+  suspectVictimContactType: null,
+}
+
+let report = {
```
danprime: Feels like this could be cloned via a deep-copy method or serialized/deserialized from the empty report.

sleepycat: Next PR!

PR #601, opened 2019-08-21
Newbackbutton

DianeLiu2019:
changed  the "button backlink" under the progress bar to "backlink" to link to previous page.


danprime:
@nmakuch Should review the backbutton

nmakuch:
I think we should change the arrow to one that takes up less space
[check out the govuk example](https://design-system.service.gov.uk/components/back-link/)

I also think we should probably make the link itself into a component as it's going to appear on every page. I could do this once she is finished the markup and refine.

PR #674, opened 2019-08-30
Datepickautoblankp2

DianeLiu2019:
date picker automatically blank for p2. victims can choose to pick a date or leave it blank. They can also type the date.


sastels:
I think we can close this PR - we've moved the date fields to just be standard text inputs rather than using the date picker.

PR #695, opened 2019-09-04
Content changes to P1 after content review

anikbrazeau:


sastels:
One thing: a button says Next: Review Report", but the next page is still "Confirm report information"

PR #713, opened 2019-09-04
French links

keithiancolbourne:
Links on final pages were going to English pages


frontend/src/p2/NextSteps.js
```diff
@@ -12,107 +13,142 @@ import { Layout } from '../components/layout'
...
+                    Report to your financial institution, if you lost money.
+                  </Trans>
+                </Li>
+                <Li>
+                  <Trans>
+                    Report to the credit bureaus,{' '}
+                    <A href="https://www.consumer.equifax.ca/personal/">
+                      Equifax
+                    </A>{' '}
+                    and <A href="https://www.transunion.ca/">TransUnion</A>, if
```
sastels: Equifax and TransUnion also have French pages.

keithiancolbourne: ah, I made a mental note about those but then forgot! Thanks Steve-o

sastels:
oh! it's failing because the test can't find the language in the graphql client. You need to do something like this
https://github.com/cds-snc/report-a-cybercrime/blob/master/frontend/src/components/topbanner/__tests__/TopBanner.test.js#L16

sastels:
:shipit: once things are passing tests 😄

PR #716, opened 2019-09-04
Replace text areas with inputs of type text

sleepycat:
P1 is done.


sastels:
Can you set the font in the input to the theme? We should also add the focusColor border.
image

vs

image

anikbrazeau:
@sleepycat 

Should the "other" checkbox inputs for P1 also be text inputs rather than text areas?
- other method of payment
- other language of communications

Should the contact information on prototype 2 also be text inputs rather than text areas?

sleepycat:
@anikbrazeau :thinking:
A fair point. I'll change that. Good catch!

PR #717, opened 2019-09-04
Remove logger

sleepycat:
This library logs to stackdriver, which is already happening when this is run
on GKE.


sastels:
True for AKS too?

sleepycat:
Shipping our logs to Stackdriver is even less desirable on AKS. These providers scoop up stdout and "do the right thing" for their platform. Good old `console.log` is the way to go.

PR #760, opened 2019-09-23
Update frontend deps

sleepycat:
As advertised.


sastels:
Moving to new frontend dir, don't need.

PR #766, opened 2019-09-25
Take down scammmmmerz 

anikbrazeau:


keithiancolbourne:
We don't need this as my confirmation tests are to clean this page up. Thanks though!

PR #819, opened 2019-10-17
Switch back to gcr

sleepycat:


keithiancolbourne:
Can we please hold off on this? I am hoping to connect with MS to get Flux going today. I would really really like to not have to go back to Google. There's no emergency to have the site being updated, right?
@sleepycat

PR #824, opened 2019-10-17
Update flux

sleepycat:
This commit updates flux to 1.15.0 and adds it as a resource to the
kustomization file for aks.

### DIFF
--- a/platform/overlays/aks/flux.yaml
+++ b/platform/overlays/aks/flux.yaml
@@ -4,15 +4,14 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  namespace: flux
   labels:
     name: flux
   name: flux
+  namespace: flux
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
 metadata:
-  namespace: flux
   labels:
     name: flux
   name: flux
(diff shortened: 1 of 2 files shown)


platform/overlays/aks/flux.yaml
```diff
@@ -54,7 +52,7 @@ spec:
   template:
     metadata:
       annotations:
-        prometheus.io.port: "3031" # tell prometheus to scrape /metrics endpoint's port.
+        prometheus.io/port: "3031" # tell prometheus to scrape /metrics endpoint's port.
```
sastels: / instead of . ?

platform/overlays/aks/flux.yaml
```diff
@@ -69,18 +67,27 @@ spec:
...
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
+        readinessProbe:
+          httpGet:
+            port: 3030
+            path: /api/flux/v6/identity.pub
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
         volumeMounts:
-        - name: acr-auth
```
sastels: What was this for?

sleepycat: This was part of the authentication stuff for Azure Container Registry. It doesn't apply for the moment so I deleted it. Will revisit soon.

Issue #749, opened 2019-09-19
Change English strings to be ID keys


anikbrazeau:

sastels:
split into individual issues.

PR #837, opened 2019-10-22
migrate lingui ids in landing page to new process

sastels:
re #749 

change the lingui ids to describe what the string is rather than being the English value of the string. This will allow us to change the English and French content more easily.

In code such as
```
<Trans>Report a scam</Trans>
```
I specified what key to add by giving it as a prop:
```
<Trans id="landingPage.title">Report a scam</Trans>
```

Also temporarily made `npm run extract` not do the cleaning (as we want to be able to extract without cleaning when doing this migration).


sastels:
[Refactor app to use IDs as keys for Lingui instead of strings](https://trello.com/c/QKb5tUtv/27-refactor-app-to-use-ids-as-keys-for-lingui-instead-of-strings)

PR #847, opened 2019-10-23
use IDS as key for Lingui instead of strings for ScammerDetail Page

DianeLiu2019:
use IDS as key for Lingui instead of strings to make easier for changing string in Scammer detail page


sastels:
only thing I notices was that the *bold* text wasn't coming through in the French page:
English:
image

French:
image

frontend/locale/fr/messages.json
```diff
@@ -184,6 +184,19 @@
...
+  "scammerDetailPage.content1": "Pensez à inclure des choses telles que:",
+  "scammerDetailPage.content2": "Qui l'escroc a prétendu être",
+  "scammerDetailPage.content3": "Où ils vous ont demandé d'envoyer des choses",
+  "scammerDetailPage.content4": "Quelle langue ils utilisaient pour communiquer",
+  "scammerDetailPage.content5": "Des captures d'écran, des messages ou des reçus",
+  "scammerDetailPage.describeFileContent": "Décrivez ce que ce fichier montre",
+  "scammerDetailPage.hint": "N'oubliez pas d'inclure vos adresses électroniques, numéros de téléphone ou liens de sites Web",
+  "scammerDetailPage.intro": "Tout indice sur la source de l’escroquerie peut aider la police à retrouver l’escroc",
+  "scammerDetailPage.removeFileButton": "Effacer le fichier",
+  "scammerDetailPage.summary": "<0> Que savez-vous sur l'origine de l'arnaque? </ 0>",
```
sastels: I think the `</ 0>` here should not have that space, ie be `</0>`

frontend/src/p2/LandingPage.js
```diff
@@ -50,7 +50,7 @@ export const LandingPage = props => {
         <Trans id="landingPage.intro">
           Scams can happen to anyone. They’re not always easy to recognize and
           new ones are invented every day.
-        </Trans>
+        </Trans>..
```
sastels: The `..` shouldn't be here

PR #865, opened 2019-10-25
change to use ID as key for lingui instead of string in contactinfo page

DianeLiu2019:
change to use ID as key   for Lingui instead of string in contactinfo page, make to change the string easier


sastels:
*sigh* I can't access the Azure test results because I don't have access to RCMP Azure, and the Cloudbuild tests didn't run because ..... (?)  We really should be doing better than this.

PR #866, opened 2019-10-25
change to use ID as key instead of string for lingui in confirmation

DianeLiu2019:
change to use ID as key instead of string for lingui in confirmation, make to edit string easier


frontend/locale/fr/messages.json
```diff
@@ -171,6 +171,17 @@
   "You will be asked to:": "Vous pourrez :",
   "You're not the only one affected by this scam. Help protect others by telling us how the scam affected you.": "Vous n'êtes pas la seule personne affectée par cette fraude. Aidez-nous à protéger les autres en nous expliquant comment la fraude vous a affecté.",
   "Your reference number is <0>#NC300234234</0>. Keep this number for your records or to add more information later.": "Votre numéro de référence est le <0>#NC300234234</0>. Conservez ce numéro dans vos dossiers pour ajouter des informations plus tard.",
+  "confirmationPage.ImpactTitle": "Impact",
```
sastels: I think the French here should be "Répercussions"

DianeLiu2019: Changed .I use online translator, it is not accurate sometimes.

PR #870, opened 2019-10-28
Add v2 of the frontend service

sleepycat:
This slimmed down frontend has had p1 removed and been replatformed on top of
Create React App instead of Razzle. It also uses React Router instead of Reach
Router in anticipation of their impending merger.


f2/src/components/buttons-container/_tests_/buttons-container.test.js
```diff
@@ -0,0 +1,107 @@
...
+  it('properly renders landing page button', () => {
+    const { getAllByText } = render(
+      <MemoryRouter initialEntries={['/']}>
+        <ThemeProvider theme={theme}>
+          <I18nProvider i18n={i18n}>
+            <ButtonsContainer
+              buttonLink={true}
+              cancel={false}
+              landing={true}
+              route="/p1/scaminfo"
```
sastels: lol @ old paths randomly hardcoded into tests

PR #892, opened 2019-10-30
Add github action to test f2

sastels:


sastels:
> To facilitate handoff this should be implemented as an Azure pipeline instead

Agreed, but handoff is not for a while - Right now testing in Azure is mostly useless since only RCMP staff is allowed to see the results.

sastels:
checked with @keithiancolbourne , he's ok with testing this way until we get full access to Azure.

PR #896, opened 2019-10-31
Ignore .env files in the platform folder

sleepycat:
As advertised

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,5 @@ node_modules
 .idea
 build-sizes.json
 data
-**/.env
 f2/build
-**/.env
+platform/**/.env.*


.gitignore
```diff
@@ -4,6 +4,5 @@ node_modules
 .idea
 build-sizes.json
 data
-**/.env
 f2/build
-**/.env
+platform/**/.env.*
```
sastels: We probably want to have .env.arango.example file, for example, in github though. Any way to include them but exclude the rest?

sleepycat: Those .env.example files are used by [dotenv-safe](https://www.npmjs.com/package/dotenv-safe) to ensure that we have loud, descriptive failures when env vars are missing.
We will still have them in other folders but nothing uses those in the platform folder.

PR #904, opened 2019-11-04
port what happened page to f2

### LINKED ISSUE #874: Move whathappened page to f2

sastels:
resolves #874


sastels:
> We should remove the content of those Trans tags. After the first extract they aren't needed anymore.

I'm torn on this. The content isn't needed for the app to run, but having it there makes it easier to understand the code IMO. Mind you, if we change the English in the lingui files then it could be a nightmare for future devs to figure out what the heck is going on. Yeah maybe I'm coming around on this.

PR #909, opened 2019-11-05
Use skaffold for local dev

sleepycat:
This commit introduces a new dev workflow based on Minikube and skaffold.
Assuming Minikube, skaffold and kustomize are installed, all you should have to
do to get started is:

```sh
make secrets
minikube start
make dev
```

You should be able to connect to `localhost:4503` and see the running app.


sastels:
As discussed IRL we should hardcode the `DB_NAME` and `DB_USER` in `make secrets` to `cybercrime` and `root`

platform/overlays/minikube/README.md
```diff
@@ -53,6 +53,8 @@ Running `make secrets` will create the files you need in the `platform/overlays/
 
 Take a moment and fill in the values.
 
+On a Mac you will probably need to install and use a version of `make` from this decade. Run `brew install make` and then try `gmake secrets` if you have any problems.
```
sastels: 😞

sleepycat:
Nope, I diffed the config against what it was generating before.

PR #945, opened 2019-11-19
create Azure script to spin up an environment for testing

DianeLiu2019:
create Azure script to spin up an environment for testing. The Azure script  allow us to create an Azure resource group,  AKS cluster, container registry, build an image in locally and in Azure,  push the image to ACR and pull the image from ACR to AKS cluster, create an Azure storage account and  Azure storage container.  Those multiple-task can be done at one time by just running the Azure script.

### DIFF
--- a/platform/azurescript.sh
+++ b/platform/azurescript.sh
@@ -1,0 +1,28 @@
+1 ##!/usr/local/bin/bash
+  2
+  3 #Create a resource group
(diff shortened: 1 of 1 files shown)


sastels:
It looks like this PR just adds an empty file.

DianeLiu2019:
I just found it. Haha.

PR #964, opened 2019-11-28
Chakra frontend

amazingphilippe:
This PR replaces custom frontend components with [Chakra UI](https://chakra-ui.com/getting-started) components.


f2/src/components/button/__tests__/Button.test.js
```diff
@@ -1,42 +1,23 @@
...
-        <Button>
-          <em>foo</em>
-        </Button>
-      </ThemeProvider>,
-    )
-    expect(getAllByText(/foo/)).toHaveLength(1)
-  })
-
-  it('uses styling props correctly', () => {
+  it('passes WCAG color contrast', () => {
```
sastels: I'm not entirely sure what this test is testing - looks like we render the button, then compare black and white. We probably should do automated a11y testing differently than this - probably by relying more on our a11y test github action.

sastels:
You've done an incredible job here @amazingphilippe 👏 👏 💯 . There's a few linting warnings that we should fix (and also look into why linting problems are shown as warnings rather than errors 🤔 ) but I think after fixing those we should ship this PR and fixup any other little issues in followup PRs.

PR #973, opened 2019-12-04
make text fields are described once instead of twice for blind

DianeLiu2019:
make text fields are described once instead of twice for blind participant by using screen-readers.


sastels:
This is going to break another accessibility requirement, namely the association between the label and the textarea on 
https://github.com/cds-snc/report-a-cybercrime/pull/973/files#diff-5f929dbf5cd7f39c966a54d1040c0a25R30

See https://dequeuniversity.com/rules/axe/3.1/label?application=lighthouse

We could make the original label hidden to the screen reader. I dunno. We should probably check with @JuliannaR to see what's preferable.

PR #978, opened 2019-12-05
fix accessibility issue for contactPage

DianeLiu2019:
fix accessibility issue for contactPage: add format and example for full name, email address, and post code field.


f2/src/locales/en.json
```diff
@@ -26,20 +41,14 @@
...
   "contactinfoPage.title": "Leave your contact information",
-  "fileUpload.added": "File added",
-  "fileUpload.removed": "File removed",
-  "impactPage.affected1": "Device or account hacked",
-  "impactPage.affected2": "Information stolen",
-  "impactPage.affected3": "Money lost",
-  "impactPage.affected4": "Reputation damaged",
-  "impactPage.affected5": "Safety threatened",
-  "impactPage.affected6": "Wellbeing affected",
-  "impactPage.affected7": "No impact",
```
sastels: `npm run extract` deleted all these lines (and added others) when it really shouldn't have. For now we should change `en.json` and `fr.json` here to only change the `contactinfoPage...` lines. PR #985 will fix `npm run extract`'s behaviour.

PR #999, opened 2019-12-13
Added comments to pipeline yaml

ngosset:
As discussed, added some comments to pipeline yaml

### DIFF
--- a/f2/azure-pipelines.yaml
+++ b/f2/azure-pipelines.yaml
@@ -1,5 +1,6 @@
 #ncfrs pipeline
 
+# Trigger pipeline on changes to master in repos
 trigger:
   branches:
     include:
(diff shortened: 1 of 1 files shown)


f2/azure-pipelines.yaml
```diff
@@ -11,22 +12,24 @@ trigger:
...
   vmImage: 'ubuntu-16.04'
 
+# Run NPM install to install depedencies on VM
 steps:
   - task: Npm@1
     inputs:
       command: 'install'
       workingDir: 'f2'
-
+# Test npm compile and other npm scripts in Azure pipeline
```
sastels: The npm compile step is where we build the lingui files for internationalization. Poorly named now that I think of it. Maybe for the comment here just say something like “run automated tests”.

ngosset: Done

PR #1001, opened 2019-12-17
Chakra frontend

amazingphilippe:
Updated components to match changes in the wireframes

- Radio buttons should work, follow [this react final form example](https://final-form.org/docs/react-final-form/examples/chakra)
- Tweaked header and footer regions
- Tweaked input interaction


f2/src/forms/ImpactStatementInfoForm.js
```diff
@@ -81,23 +82,27 @@ export const ImpactStatementInfoForm = props => {
...
+                            key={key}
+                            name="howWereYouAffected"
+                            value={key}
+                            isChecked={checked}
+                          >
+                            {i18n._(key)}
+                          </CheckboxArrayControl>
+                          <ConditionalForm show={checked}>
+                            Conditional form
+                          </ConditionalForm>
```
sastels: Was this ConditionalForm for debugging?

amazingphilippe: Yes, I was working on it last night and I could not finish it. Did not realize that it would append to my PR. See the next commit to find working placeholder conditional forms!

We can discuss how we want to implement conditional fields. I was thinking we could place them in another file. I'm a little worried about readability... there must be a better way

sastels: looks good! only thing I'm seeing is when I load the page the fonts are initially different and then change after a fraction of a second. Any thoughts about what's doing that?

amazingphilippe: I'm loading a google font. Maybe I'm not doing it efficiently. Look for it in app.js
I was looking for a better way to load it, maybe as a promise, but could not find anything. Lets do an issue to fix this

sastels: I see. Yeah, we can push that to another PR. Probably we can just download the fonts and include them in the repo, then serve them locally.

PR #1003, opened 2019-12-18
Change local state management

sastels:
We need local state in our app to store the values on the forms before the user reaches the confirmation page. Previously we were using GraphQL since we needed GraphQL anyway (to talk to the api). We no longer have the api, and react now supports local state via hooks, so in this PR we switch to a react-hooks based approach.


f2/src/WhatHappenedPage.js
```diff
@@ -10,58 +10,61 @@ import { Steps } from './components/stepper'
...
+              <Trans id="whatHappendPage.backButton" />
+            </BackButton>
 
-            <P>
-              <Trans id="whatHappendPage.details1" />
-            </P>
+            <Stack spacing={4} role="heading" aria-level="1">
+              <Steps activeStep={2} totalSteps={6} />
+              <H1 as="span">
+                <Trans id="whatHappendPage.title" />
```
naomiseminega: Typo in 'whatHappendPage'  (should be whatHappenedPage), same typo reflected in en.json.

sastels: good catch! let's add an issue to correct this spelling - it's all over the place (cut and paste FTL)

PR #1017, opened 2019-12-19
Upgrade nodejs version in Dockerfile

### LINKED ISSUE #1016: Update nodejs in Docker container from from 12.9 to 12.14
Need someone to update nodejs in Dockerfile to 12.14 and test new container.

12.14 is the current LTS version. Snyk has found low severity vulnerabilities in 12.9

ngosset:
Ran docker build & docker run on local machine and clicked through application. Doesn't seem to break anything?

### DIFF
--- a/f2/Dockerfile
+++ b/f2/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.9-alpine AS build-env
+FROM node:12.14-alpine AS build-env
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


ngosset:
> It would probably be good to change the node version on line 1 as well (you've changed the version that's being used in the image, but it might be confusing to use another one for building the image)

Whoops!

Issue #1013, opened 2019-12-18
P3 contact


sastels:
https://www.figma.com/file/hNqLtiygFfxJ7GdPmGgVyO/Sprint-2?node-id=567%3A487

<img width="227" alt="image" src="https://user-images.githubusercontent.com/8228248/71102069-ceac8080-2185-11ea-8660-eb4d3c16b192.png">

PR #1020, opened 2019-12-30
New contact info form #1013

ngosset:
Updated contact page according to new figma boards

Some work still needs to be done on the Next & Back buttons when we tie these all together.

No translations yet for new french labels


ngosset:
Closing PR without merging - want to make some additional changes.

PR #1056, opened 2020-01-09
Remove currently unused import that breaks the build

taowa:
My previous commit (https://github.com/cds-snc/report-a-cybercrime/commit/dbfe3696be96aaaef20eade4b55d5fc0f225398b) fails to build. This should fix it.


sastels:
Ah, I had to pull in `master` to resolve merge conflicts in #1053 so I had to fix this there anyway. 👍

PR #1174, opened 2020-01-17
Change to use Service Principle for ACR

davidheerema:
I wanted to update this script to use a service principle to connect the app service to the ACR.  Considering this script had the admin credentials for the ACR removed a while ago, I've basically left it broken until now (theoretically)


f2/utils/azurescript
```diff
@@ -1,10 +1,11 @@
  export PROJECT_NAME=rcmpcybercrime
  export RG_NAME=MpSCCDSCybercrimeRG
  export ACR_NAME=MpSCCDSCybercrimeacr
- export IMAGE_NAME=cybercrimeimage
+ export IMAGE_NAME=f2
```
sastels: at some point we should rename our "f2" to "frontend". Have to remember to change image names at that time.

davidheerema: Yes, let me know when f2 is renamed to "frontend" so that we can reconfigure a) snyk b) the container settings for the app service.

PR #1179, opened 2020-01-20
[WIP] - added the missing configuration to the local dev a11y linter

khalidelaggan:
# Description

> Local a11y linter isn't plugged in by default with our Create React App setup

Fixes # (issue)
This adds back the linter in place, and it currently logs all issues now for each component markup in the console 

Screen Shot 2020-01-14 at 1 45 33 PM

### DIFF
--- a/f2/package-lock.json
+++ b/f2/package-lock.json
@@ -5076,7 +5076,8 @@
             },
             "ansi-regex": {
               "version": "2.1.1",
-              "bundled": true
+              "bundled": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


khalidelaggan:
This should be ready to merge

khalidelaggan:
@sastels This is now ready to be merged

sastels:
fixes #1180

f2/src/components/text-area/__tests__/TextArea.test.js
```diff
@@ -2,23 +2,10 @@ import React from 'react'
...
-      <form>
-        <label>
-          Test
-          <TextArea />
-        </label>
-      </form>,
-    ).find('textarea')
-    expect(wrapper.is('textarea')).toBeTruthy()
-  })
-
```
sastels: I think these two tests will now pass with the other changes you’ve made - can you add them back in and see?

khalidelaggan:
@sastels  this is also ready to be 🚢

sastels:
> @sastels this is also ready to be 🚢

As discussed above, I think you can restore those two deleted tests, then we should be ready to go! :)

PR #1200, opened 2020-01-23
Breadcrumbs

anikbrazeau:
# Fixing breadcrumb content

-- Making it clear that back link text is the title with quotation marks
-- Using French-style « guillemets » in French


sastels:
Missed the What Could Be Affected page
image

PR #1218, opened 2020-01-27
Lingui tests

### LINKED ISSUE #1210: Add lingui testing
## Summary

Write tests to check that all lingui keys have non-empty translations.

## Motivation

We currently do not have any tests for missing translations of lingui keys. This makes it likely that at some point we will ship a product with missing content. We need tests in our CI process to catch this.

## Drawbacks

none

## Unresolved questions

Should also add a husky hook to run `npm run extract` before commit

sastels:
# Description

Essentially put back the test we used to have where we ensured that we had English and French strings for all our lingui keys. This test was removed when we refactored how we do lingui.

Fixes #1210 

- [x] Technical debt

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made any needed changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

# Required followup work

We might still miss keys if lingui isn't picking them up (ie via `npm run extract`). We should add a precommit hook to run `npm run extract`, this will hopefully help expose this error to authors and reviewers.


f2/src/locales/en.json
```diff
@@ -222,7 +222,7 @@
   "scammerDetail.fileDescription": "Describe what this file shows",
   "scammerDetail.intro": "Any clues about the source of the scam can help police track down the scammer.",
   "scammerDetail.nextButton": "Next: Impact of scam",
-  "scammerDetail.nextPage": "scammerDetail.nextPage",
+  "scammerDetail.nextPage": " ",
```
sastels: This is old p2 code. Just changed this to pass the tests, This line will be deleted when we clean up the repo.

PR #1278, opened 2020-02-04
Update Azure Script for Log Analytics

davidheerema:
Adding a line to enable https-only and re-direct port 80 to 443 automatically for the web app.  The WAF already only listens on 443 so we're good there.

I tested the command individually and it works fine.


davidheerema:
I have added some lines to configure Log Analytics and enable some https-only settings.
I have also cleaned up the script a bit adding newlines (a steve request :)

PR #1288, opened 2020-02-05
change language switcher to a pseudobox

### LINKED ISSUE #1266: Can't tab to language toggle
## Summary

Can't get to language toggle with keyboard navigation

## Steps to reproduce

tab through any page.

sastels:
Fixes #1266

# Description

Change language switcher from a `Link` to a `PseudoBox` so that it's tabable.

# Checklist:


amazingphilippe:
We're still running into the issue where we are making a button look like a link. Would it be possible to generate a path to pass as a href to a link component? While only changing the language. This might also help with translating url paths

sastels:
> We're still running into the issue where we are making a button look like a link. Would it be possible to generate a path to pass as a href to a link component? While only changing the language. This might also help with translating url paths

AFAIK with lingui there's no way to indicate the language with a url (ie, say, by adding something like `?lang=fr` to the url) so we need to use an `onClick`. 

I think the issue right now in master is that we have a link with just an onClick and no href, and this makes it not in the navigation flow. Also... strange to have a link with no href.

amazingphilippe:
This is using react-intl, which we are not using. I wonder if its still achievable. Might also be of interest in the future.
https://dev.to/prototyp/multi-language-routing-in-react-k9l

For now, this will do.

PR #1325, opened 2020-02-07
Fix focus

### LINKED ISSUE #1142: Page focus is not consistently at the top of pages
## Summary

As you click through the form, the focus loads the page at different points rather than consistently at the top. This is especially noticeable on longer pages such as the money, information, device and business pages as well as on the suspect tinfo page.

sastels:
Fixes #1142 

# Description

When a new screen loads, scroll to the top. Previously the focus was at the middle of the new screen.

There's more lines of code here fixing the tests than fixing the app 🤦‍♂ 

# Checklist:

### DIFF
--- a/f2/src/__tests__/BusinessPage.test.js
+++ b/f2/src/__tests__/BusinessPage.test.js
@@ -13,19 +13,20 @@ i18n.load('en', { en })
 i18n.activate('en')
 
 describe('<BusinessPage />', () => {
-    afterEach(cleanup)
+  beforeEach(() => (global.scrollTo = jest.fn()))
+  afterEach(cleanup)
 
-    it('renders', () => {
-        render(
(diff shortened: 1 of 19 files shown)


f2/src/__tests__/BusinessPage.test.js
```diff
@@ -13,19 +13,20 @@ i18n.load('en', { en })
 i18n.activate('en')
 
 describe('<BusinessPage />', () => {
-    afterEach(cleanup)
+  beforeEach(() => (global.scrollTo = jest.fn()))
```
sastels: The `window` object that the app uses only exists when running in a browser, not when running in jest. So for the tests to use `window.scrollTo()`, we make the function manually by setting it to a fake jest function. And for some reason we set `global.scrollTo` rather than `window.scrollTo` 🤷‍♂

PR #1339, opened 2020-02-10
Scan file uploads for malware

### LINKED ISSUE #1219: Scan file uploads for malware
## Summary

Need to scan file uploads for malware.
Pass the filestream of the file upload to clamav and analyze the response.
Reject or ignore upload if malware identified.
See @ngosset for sample code using the clamav API.

## Motivation

We don't want users uploading malware

## Drawbacks

May cause people to call the call centre when they can't upload a file.

There may be another solution we should use instead of clamav.

## Unresolved questions

Probably need to create a separate issue for notifying the user when the upload was rejected.
Do we want to scan on the evidence page, or wait for the final submit on the confirmation page?
Possible pre-requisites:
Issue #1216 
Issue #1217

kevinyang06:
Fixes #1219 

# Description

Need to scan file uploads for malware.

# Any new packages installed?

formidable and clamd

# Checklist:

# Required followup work


f2/server.js
```diff
@@ -4,6 +4,13 @@ const path = require('path')
 const formidable = require('formidable')
 const { getAllCerts, encryptAndSend } = require('./src/utils/encryptedEmail')
 const { selfHarmWordsScan } = require('./utils/selfHarmWordsScan')
+var clamd = require('clamdjs')
+var fs = require('fs')
+var scanner = clamd.createScanner(
+  'mpsccdscybercrimeclamav.canadacentral.azurecontainer.io',
```
sastels: If we move this `scanner =...` stuff to below the `require('dotenv').config()` then we can use an env var for the url, i.e. something like `process.env.CLAM_URL`

f2/server.js
```diff
@@ -61,6 +68,17 @@ const uploadData = (req, res) => {
     console.log('Files', files)
     for (const file of Object.entries(files)) {
       console.log(file)
+      //scan file for virus
+      var readStream = fs.createReadStream(file[1].path)
+      scanner
+        .scanStream(readStream, 10000)
```
sastels: Is the 10000 some sort of timeout? If so maybe just add a comment here.

kevinyang06: fixed

f2/server.js
kevinyang06: done

f2/server.js
```diff
@@ -4,8 +4,13 @@ const path = require('path')
 const formidable = require('formidable')
 const { getAllCerts, encryptAndSend } = require('./src/utils/encryptedEmail')
 const { selfHarmWordsScan } = require('./utils/selfHarmWordsScan')
+var clamd = require('clamdjs')
+var fs = require('fs')
+
+var http = require('http')
```
sastels: we can delete this line, we don't use the `http`module in `server.js`

kevinyang06: done

PR #1365, opened 2020-02-12
Stop randomizing user info

### LINKED ISSUE #1344: Stop randomizing user info
## Summary

We need to save the real info!

## Design Detail

Strip out all the `randomizeString()` code (`server.js` and `ConfirmationPage.js`)

kevinyang06:
Fixes #1344 

# Description

We need to save the real info!

# Any new packages installed?

no

# Checklist:

# Required followup work


f2/server.js
```diff
@@ -43,14 +43,6 @@ const randLetter = () => {
 }
```
sastels: you can delete the `randLetter` and `randDigit` functions here too.

f2/src/ConfirmationPage.js
```diff
@@ -53,9 +39,6 @@ async function postData(url = '', data = {}) {
 const prepFormData = (formData, language) => {
   let contactInfo = formData.contactInfo ? formData.contactInfo : {}
   let { fullName, email, postalCode } = contactInfo
```
sastels: You can also delete these two  lines since we don't have to mess with `contactInfo` anymore, as well as the 
```
contactInfo: {
  fullName,
  email,
  postalCode,
},
```
block in the return statement.
(it looks like this was also preventing the phone number from being submitted)

kevinyang06: done

PR #1375, opened 2020-02-13
Updating content on What was affected page

anikbrazeau:
Updating content on What was affected page

# Checklist:


f2/src/forms/WhatWasAffectedForm.js
```diff
@@ -145,10 +145,6 @@ export const WhatWasAffectedForm = props => {
                 </Alert>
               </Control>
             ) : null}
-
-            <P>
```
kevinyang06: please remove the import {p} on the top since it is not used.

anikbrazeau: thanks! change made!

PR #1389, opened 2020-02-14
fix server.js

sastels:
# Description

My fix to a merge conflict caused server.js to crash when I added the Notify environment variables 😞 . This fixes that.

# Checklist:


davidheerema:
It looks like the change to server.js is just removing the randomizing of the email string that is never referenced again, so why does this break the server?  Thanks for explaining Steve.

sastels:
> It looks like the change to server.js is just removing the randomizing of the email string that is never referenced again, so why does this break the server? Thanks for explaining Steve.

The problem is that the function it’s trying to use no longer exists 😕 (we deleted it yesterday when we stopped randomizing the personal info)

davidheerema:
Ah, I see - thanks steve.

PR #1395, opened 2020-02-14
Set user language based on host

### LINKED ISSUE #1324: Set user language based on host
## Summary

CAFC will redirect to different hosts depending on the language of the user.
So we need some logic when a user first loads the react app:

If host == report.antifraudcentre.ca set locale to english
else if host == signalment.centreantifraude.ca set locale to french
else set to browser's locale

## Motivation

During beta all our traffic will come from antifraudcentre.ca redirects. Users will have already determined their language by this point and this should stay consistent when redirected to our pilot.

## Unresolved questions

Should we also add query parameter support? Example: url/?lang=fr

kevinyang06:
Fixes #1324

# Description

CAFC will redirect to different hosts depending on the language of the user.
So we need some logic when a user first loads the react app:

# Any new packages installed?

# Checklist:

# Required followup work

I couldn't test it in localhost


ngosset:
@sastels @kevinyang06 

How should we test this? Can't really test it until after we merge?

f2/src/LandingPage.js
```diff
@@ -15,6 +15,17 @@ import { TrackPageViews } from './TrackPageViews'
...
+
+if (
+  window.location.hostname ===
+  'https://signalez-un-crime-informatique.cds-snc.ca/'
+) {
+  i18n.activate('fr')
+} else if (window.location.hostname === 'report-a-scam.cds-snc.ca') {
+  i18n.activate('en')
+} else {
+}
```
sastels: We don't need the `else {}` here

sastels: Also, on Chrome in the console I'm just getting `signalez-un-crime-informatique.cds-snc.ca` for the hostname:
image
It might be better to do something like
```
if (window.location.hostname .indexOf('signalez')> -1) ....
else if (window.location.hostname .indexOf('report')> -1) ...

sastels:
> @sastels @kevinyang06
> 
> How should we test this? Can't really test it until after we merge?

We could probably write a test that mocked the hostname, but IMO in this case it's probably sufficient to merge and ensure it works.

PR #1514, opened 2020-02-22
fix for skipping location

### LINKED ISSUE #1513: skipping the location page breaks the analyst report code
## Summary

If the location is skipped then `location.postalCode` is not submitted, which causes the server's report code to exit early.

## Steps to reproduce

1) skip the location page
2) enter your email
3) the intake report you get says "ERROR in formatAnalystEmail..."

sastels:
Fixes #1513 

# Description

Save location as `{postalCode: ''}` even if skipped.

# Checklist:


sastels:
> Seems good. Do we have the same problem with the contact information skip button?

no, we did the fix for it a while ago (when we added the skip to that page)

PR #1524, opened 2020-02-23
Resource tagging

davidheerema:
Changes to the Azure CLI to rename resources and tag them.


davidheerema:
@sastels Yes, I recall that too and I said the same thing to Graham on Cloud Ops - he was the one who proposed these changes.  He said he checked them out, but we'll find out for sure when I run this momentarily / tomorrow.

PR #1525, opened 2020-02-23
A11y fixes

khalidelaggan:
Fixes #(issue)

- Mid form feedback Textarea label missing 
- Page  landmark 
- Changed the feedback button to slightly darker blue 
- App version for screen reader 

# Description

# Any new packages installed?

> Give details

# Checklist:

# Required followup work


f2/src/App.js
```diff
@@ -62,18 +62,19 @@ const App = () => {
             bg="gray.50"
           >
             <Home />
-          </Stack>
 
-          <Layout>
-            <P fontSize="sm" my={3}>
+            <Layout>
+
+            <P fontSize="sm" my={3} aria-label="application version">
```
khalidelaggan: Removed this to the main for a better landmark structure

khalidelaggan: @amazingphilippe  let me know if this small addition/hack to aria-label here will cause any issue. I'm trying to avoid refactor P at this point to pass aria attr as props :)

f2/src/components/button/index.js
```diff
@@ -13,7 +13,7 @@ export const Button = ({ variant, variants, variantColor, ...props }) => (
       ? {
           borderBottomWidth: '3px',
           borderBottomColor: `${variantColor}.800`,
-          bg: `${variantColor}.600`,
+          bg: `${variantColor}.700`,
```
khalidelaggan: Not sure if this is the best way to apply the color change for the feedback button in chakra, as this will also impact all non-default buttons themes. happy to discuss this further  :)

amazingphilippe: The only thing I can think of is that we are losing the 'active' state styles: when a button is mousedown. 

I see two options: 
1. Choose an even darker color for the active state
2. We make sure the text is larger than 18px. AAA minimum color contrast for that size is 4.5, which we are passing on the buttons at 5.65
Screen Shot 2020-02-24 at 8 32 47 AM

amazingphilippe: I would maybe focus more towards the footer links that are barely passing at 4.61

khalidelaggan: hmmm, little confused as the current 
Screen Shot 2020-02-24 at 11 28 46 AM
footer links and color contrast pass the required checks

khalidelaggan: will check against master now and see if we still have this issue

amazingphilippe: @khalidelaggan They all pass because they are considered large text. I say we merge this without changing the button color.

f2/src/forms/MidFeedbackForm.js
```diff
@@ -148,7 +148,7 @@ export const MidFeedbackForm = props => {
                   <Field name="problemDescription">
                     {props => (
                       <FormControl>
-                        <FormLabel>
+                        <FormLabel htmlFor="problemDescription">
```
amazingphilippe: This is fixed by #1496 We haven't merged it yet

f2/src/components/button/index.js
```diff
@@ -13,14 +13,14 @@ export const Button = ({ variant, variants, variantColor, ...props }) => (
...
         }
       : '')}
     _hover={{
       boxShadow: 'outlineHover',
     }}
     _active={{
-      bg: `${variantColor}.700`,
+      bg: `${variantColor}.800`,
```
khalidelaggan: @amazingphilippe I've added a darker shade now for the active state

f2/src/components/button/index.js
khalidelaggan: As this form is inherited in each page, I would rather fix this minor adjustment for the button color now and have us in a clean state (0 issues)reported by any a11y check tools instead.

f2/src/forms/MidFeedbackForm.js
amazingphilippe: We'll need to be careful when merging into master. We have an other PR for this file.

khalidelaggan: will remove it from this pr shortly

khalidelaggan: @amazingphilippe  this is now removed from this PR

PR #1526, opened 2020-02-23
Blob Storage and App Variables

davidheerema:
Adding the blob storage and app service environment variables


utils/azurescript.sh
```diff
@@ -32,6 +34,13 @@ export APP_SUBNET_RANGE=10.9.0.0/24
 export CONTAINER_SUBNET="${VIRUS_SCANNER_NAME}SN"
 export CONTAINER_SUBNET_RANGE=10.9.1.0/24
 
+## App Environment Variables
+export NOTIFY_API_BASE_URL=
+export NOTIFY_API_KEY=
+export NOTIFY_CONFIRMATION_TEMPLATE_ID=
+export REACT_APP_GOOGLE_ANALYTICS_ID=
+export SELF_HARM_WORDS=
+export SEND_UNENCRYPTED_REPORTS=yes
```
sastels: SEND_UNENCRYPTED_REPORTS is only for dev environments (at least, with the value “yes”)

utils/azurescript.sh
```diff
@@ -66,6 +75,10 @@ az cognitiveservices account create --name $COGNITIVE_NAME --resource-group $RG_
 ## Create Database
 az cosmosdb create --name $DB_NAME --kind MongoDB
 
+## Create Storage account for blobs
+az storage account create --name $BLOB_NAME --resource-group $RG_NAME --sku Standard_RAGRS --kind BlobStorage --access-tier Hot
+az storage account blob-service-properties update --enable-delete-retention true --delete-retention-days 100 -n $BLOB_NAME -g $RG_NAME
```
sastels: Is this going to delete the files after 100 days?

utils/azurescript.sh
```diff
@@ -32,6 +34,13 @@ export APP_SUBNET_RANGE=10.9.0.0/24
 export CONTAINER_SUBNET="${VIRUS_SCANNER_NAME}SN"
 export CONTAINER_SUBNET_RANGE=10.9.1.0/24
 
+## App Environment Variables
+export NOTIFY_API_BASE_URL=
+export NOTIFY_API_KEY=
+export NOTIFY_CONFIRMATION_TEMPLATE_ID=
+export REACT_APP_GOOGLE_ANALYTICS_ID=
```
sastels: This Google Analytics one is just needed in the docker build pipeline, not here.

utils/azurescript.sh
davidheerema: So I should set it to no?

utils/azurescript.sh
davidheerema: This is enabling "soft delete" so files will hang around invisibly if deleted for 100 days
I can put it to a max of 365, but without this setting "soft delete" is OFF by default

PR #1539, opened 2020-02-24
Removed <Report ID> from thank you page. Replaced with conditional Report Id display.

### LINKED ISSUE #1472: <ReportID> is displayed on thank you page
## Summary

<ReportID> is displayed on thank you page

error

## Steps to reproduce

## Unresolved questions

naomiseminega:
Fixes #1472

# Description
- Removed <Report ID> from thank you page
- Replaced with conditional Report Id display
- As per recommendations in Issue #1472:
    * When report ID starts with "NCFRS" msg is displayed that report submission was not successful
    * Otherwise , the generated report Id is displayed.
    *Also, figured we didn't need a line that says 'Submitting your report', since by the time there is a 'Thank You' page, the report has been submitted (it was misleading (?)).

# Required follow up work
- I noticed from the code that all report Id's that are generated start with NCFRS (?), not sure if this is a placeholder or will stay this way. We may have to update the logic in that case.

# Any new packages installed?
No
# Checklist:


f2/src/forms/__tests__/HowDidItStartForm.test.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
```
sastels: Isn't this test already in the code?

sastels:
review app deployed at https://rac-prototyp-report-id-vaicqvy.herokuapp.com/

f2/src/ThankYouPage.js
```diff
@@ -54,17 +54,23 @@ export const ThankYouPage = () => {
...
+                  </Text>
+                </Trans>
+              ) : (
+                <Trans
+                  id="thankYouPage.referenceNumber"
+                  values={{ reference: state.formData.reportId }}
+                >
+                  <Text as="span" d="block" />
+                </Trans>
+              )}
```
sastels: * We should flip the order of these two blocks (ie show the reportId when `state.formData.reportId && state.formData.reportId.startsWith('NCFRS-')`)
* also, we should have a third option... something like when the confirmation page loads (ie in `ConfirmationPage.js` after the `const { i18n } = useLingui()`)
```
if formdata.reportId !== '') 
 dispatch({ type: 'saveFormData', data: { reportId: '' } })
```
 to make the reportId an empty string. Then on the thankyou page we could also check for `reportId ===''` and if so show something like "Submitting report..." (which would then get replaced by the number or the error message when there's a response from the server)

naomiseminega: Sounds good! I'm on it

f2/src/forms/__tests__/HowDidItStartForm.test.js
naomiseminega: Yes it is! Oups, not sure how it ended up here again :P

f2/src/ThankYouPage.js
naomiseminega: @sastels fixed! Hopefully I covered all the cases :)

f2/src/ThankYouPage.js
```diff
@@ -29,6 +29,22 @@ export const ThankYouPage = () => {
 
   const [, dispatch] = useStateValue()
 
+  // Message displayed on Thank you Page
+  let reportId = state.formData.reportId
+  let thankYouMessage
+
+  if (!reportId || reportId === '') {
+    thankYouMessage = <Trans id="thankYouPage.reportSubmission" />
```
sastels: I didn't know you could do this. Nice! 🎉

naomiseminega: Me either! Lots of googling :p :p

Issue #1421, opened 2020-02-17
Bug: Server error gets printed in the 'reference number field' on the Thank You Page


naomiseminega:
## Summary
If you go through the App (whether running on server or dev), and submit the report, the latest server error will be printed in green box at the top of the Thank you Page after the 'Your reference number is ' field.

## Steps to reproduce
- Navigate to localhost:3000
- Go through the App (adding dummy info is optional)
- Submit report, and check the green box at the top of the Thank you page 
- Do Fn F12 and compare the printed message to the most recent server error.

## Unresolved questions

> Are there any related issues you consider out of scope for this issue that could be addressed in the future?

naomiseminega:
<img width="1386" alt="Screen Shot 2020-02-17 at 12 44 42 PM" src="https://user-images.githubusercontent.com/58989481/74676754-bb696180-5184-11ea-999e-1a579fa8b0e1.png">

naomiseminega:
@sastels  I can close this bug right? It will be fixed by Norm's upcoming PR. Seems like the same bug Emily got.

sastels:
Yeah, I think so. 👍

PR #1559, opened 2020-02-25
(Leave for after Launch) Server file upload - size validation

### LINKED ISSUE #1256: Restrict file upload by file extension and size.
## Summary

We are using FRS as a baseline for file upload.
We should have the same upload restrictions: jpg, jpeg, doc, docx, xls xlsx, pdf, txt, rtf. File size restriction: 4MB

## Motivation

Limiting the amount of stuff we can receive at the start. We can iterate on this and adjust as needed.

## Unresolved questions
FRS also limits a user to 3 uploads per complaint, but is this necessary? 
How many 4MB files can we fit in an e-mail through RCMP mail gateway?
OCRE allows pngs, I don't think it would hurt to add pngs to start?

naomiseminega:
Fixes #1256 

# Description
-Added file size validation on Server side, to catch any file bigger than 4MB
- Server won't allow saving the file if it's bigger than 4MB limit
- It will log warning

# Any new packages installed?
No
> Give details

# Checklist:

# Required followup work
No


f2/server.js
```diff
@@ -54,10 +58,13 @@ const uploadData = async (req, res, fields, files) => {
...
-  // Await here because we also need these results before saving
-  await scanFiles(data)
+  // Check file size
+  if (checkFileSize(files)) {
+    // Await here because we also need these results before saving
+    await scanFiles(data)
 
-  contentModeratorFiles(data, () => save(data, res))
+    contentModeratorFiles(data, () => save(data, res))
+  }
```
sastels: Running locally It hangs if I add a file - I think we need some sort of 
```
else {
  res.statusCode = 502
   res.statusMessage = 'Error: files too big'
   res.send(res.statusMessage)
}
```
here

naomiseminega: Fixed!

sastels:
Running locally, if I do basically a blank report without files I get 
```
Your reference number:
Error: files too big
```

naomiseminega:
> Running locally, if I do basically a blank report without files I get
> 
> ```
> Your reference number:
> Error: files too big
> ```
> 
> oh! I think it's because `CheckFileSize()` returns true if and only if the files are too big, and in `server.js` we scan and save if `CheckFileSize()` is true.
> 
> Probably best to rename the function something like `FilesAreTooBig()` or something to make it clear what `true` and `false` mean 😄

Oh :p ! I remember I got the same bug last time (remember issue  #1421 (?), server error was being printed on the thank you page. I'll try your suggestion.

sastels:
Will resolve in #1648 instead (using a different approach since we now are dealing with files slightly differently on the server)

PR #1593, opened 2020-02-28
Hookup finalfeedback thanks page

### LINKED ISSUE #1592: Confuse info on the thank you page when the victim cancels the report and gives feedback
## Summary

When the victim cancels the report and fills out the feed back, it should go back to "Thank you for trying our new website" page and with an extra sentence "Thank you. We have received your feedback." instead of "Thank you for reporting" with sentence  "Submitting your report…" because the victim has not submitted any report yet

## Steps to reproduce

- Fill out a few pages of report
- Click "Cancel"
- Click "Cancel Report"
- Click " Tell us more"
- Fill out the form and click submit

cancel

## Unresolved questions

sastels:
Fixes #1592 

# Description

* submitting final feedback (from Cancel or thanks page) leads to finalfeedbackthanks page
* also took out the back button on the finalfeedback page so it'd work with both cancel and thanks

# Checklist:

# Required followup work

We should probably just make our Back buttons do a `history.back()`. Though then they're exactly the same as pressing the browser back button, so is there a point?

### DIFF
--- a/f2/src/FinalFeedbackPage.js
+++ b/f2/src/FinalFeedbackPage.js
@@ -7,7 +7,6 @@ import { H1 } from './components/header'
 import { TrackPageViews } from './TrackPageViews'
 import { Layout } from './components/layout'
 import { FinalFeedbackForm } from './forms/FinalFeedbackForm'
-import { BackButton } from './components/backbutton'
 import { Stack } from '@chakra-ui/core'
 import { useStateValue } from './utils/state'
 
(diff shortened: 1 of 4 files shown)


f2/src/FinalFeedbackPage.js
```diff
@@ -42,17 +41,14 @@ export const FinalFeedbackPage = () => {
...
-            </BackButton>
             <H1>
               <Trans id="finalFeedback.title" />
             </H1>
             <FinalFeedbackForm
               onSubmit={data => {
                 submitToServer(data)
                 setState((state.doneFinalFeedback = true))
-                history.push('/thankYouPage')
+                history.push('/finalfeedbackthanks')
```
amazingphilippe: Can we add a conditional here that links to finalfeedbackthanks only if state.doneforms is true? That way, we keep what we have now, but we fix the confusing scenario described in #1592

sastels: added code so that
- if they've submitted, they go back to the thankyou page that has the report id on it
- if they have cancelled it goes to the "Thank you for helping us improve this service!" page

PR #1597, opened 2020-02-28
Updating Azurescript - IMAGE_NAME

davidheerema:
Small and simple improvement from my backlog of bugs.  Probably lowest priority possible.


davidheerema:
If you really wish to test it you can request a sandbox azure resource
group that you can run the script against I guess, then delete the group
after.

Or, just trust me, until we get this script deprecated and use some
infrastructure as code.

PR #1620, opened 2020-03-03
Storybook

amazingphilippe:
# Description

This is for component documentation. As we have one test per component, we should have a story for each component. This will ensure we use components correctly across the app. 

`npm install` then `npm run storybook` to run storybook locally. it will be available at http://localhost:9009

# Any new packages installed?

Yes. Mainly stuff related to storybook itself and its dependencies, and its presets for create-react-app

# Checklist:

# Required followup work

1. Create stories for components!
2. Can we fix lingui? It looks a bit wonky when we have translations in the components


sastels:
Discussion on using liungui and storybook
https://github.com/lingui/js-lingui/issues/413

PR #1628, opened 2020-03-05
Combine analyst emails

### LINKED ISSUE #1450: Reformat email sent to analyst
## Summary

Using HTML to format data into more readable format with headings, spacing, bolding, etc.

Transforming the date format to YYYY-MM-DD time so that it's easier to input into existingCAFC FRS system.

## Design Detail

Current format: 

Screen Shot 2020-02-19 at 2 02 30 PM

### LINKED ISSUE #1579: Send files and report as a single e-mail
## Summary

Files are sent as a separate e-mail. This is a problem as files can come in out of order and it's not clear which report they are from. We should try to send everything as one e-mail.

## Motivation

This is also important because an offensive image could be received by the analyst before they see the report that has a warning image might be offensive

## Unresolved questions

Additionally: Can we make the warning about potential offensive image more apparent (bigger font or part of the subject?)

### LINKED ISSUE #1652: (Outlook) The French text in analyst report is not displayed correctly with special characters
## Summary

In outlook, The French text in analyst report is not displayed correctly with special characters 
Here is example

What happened:           B+Ce n'est pas normal. C'est une anomalie de ne pas considC)rer les femmes pour des postes importants. Moi, j'aimerais entendre ces associations sportives venir de l'avant et nous expliquer le pourquoi. Il y a peut-C*tre une bonne raison qu'on ne connaC.t pas, mais dites-nous-la et on va finir par comprendre. Mais je ne pense pas qu'il y en ait uneB;, affirme la rC)putC)e femme d'affaires quC)bC)coise DaniC(le Henkel, qui a dC; trimer dur pour percer l'univers masculin de l'entrepreneuriat avant d'en devenir une chef de file

## Steps to reproduce

## Unresolved questions

sastels:
Fixes #1652 
Fixes #1579 
Fixes #1450 

# Description

Use `nodemailer` library to combine files and report into a single email. This means we don't have to use `makemime` anymore.

Also, if `LDAP_UID` and `MAIL_TO` are empty then the unencrypted report will be emailed to the user if `contactinfo.email` is filled in.

# Any new packages installed?

no, we were already using `nodemailer` to connect to the SMTP server.

# Checklist:


ngosset:
Mind you the problem I had above could be because content moderator failed?

f2/src/utils/encryptedEmail.js
```diff
@@ -55,75 +57,75 @@ const getCert = uid => {
...
 
+const prepareUnencryptedReportEmail = (message, data, callback) => {
+  let transporter = nodemailer.createTransport({
+    streamTransport: true,
+    newline: 'unix',
+    buffer: true,
+  })
+
+  let attachments = data.evidence.files
+    .filter(file => !isProductionSystem || file.malwareIsClean)
```
sastels: right now we're not connecting to Clam outside prod - just a hack so that we'll get files attached anyway.

sastels: got our own clam and removed this hack.

kevinyang06:
image
I'm not able to received files attachment on outlook

sastels:
> I'm not able to received files attachment on outlook

I'm not sure why the file info table is messed up ☝️  but (I think) the reason you didn't get a file attached is that the virus scanner didn't run (so the system defaults to assuming the file has a malware and doesn't attach). You might need a new `CLAM_URL`

PR #1656, opened 2020-03-10
Prod version number

### LINKED ISSUE #1444: Add a prod version environment variable
## Summary

Have a sensible version number for prod

## Design Detail

Add `REACT_APP_PROD_VERSION`
Use this in the version along with the gitSha. Something like
Version: 1.0.0 (a2ef182)

## Unresolved questions

Shall we stick to semantic versioning? What would be a major version in that case? Some sort of significant milestone?

sastels:
Fixes #1444 

# Description

Add a version number of the form 1.0.0 (in addition to the gitsha that is currently there).

# Checklist:


sastels:
> This looks good to me Steve.
> 
> **Please inform me when this is merged so that I can add the new environment variable to every environment**

You shouldn't have to add anything for this - at least, the way I've done it here the values are hard coded in everywhere (and are just needed in the pipeline / docker build, not in App Service)

davidheerema:
Oh yes, I see that.  My mistake thanks.

sastels:
I think this is probably the wrong way to do this...  if we're going to hard code the version # anyway, we should just hard code it in the app code directly.

I shall fix up...

PR #1682, opened 2020-03-17
Get email addresses from ldap query

### LINKED ISSUE #1633: Out of order HRMIS IDs will encrypt for the wrong email.
Regarding the environment variables:

> comma delimited list of HRMIS should be same order as comma delimited list of e-mail
- Norm

If they are out of order then people will receive emails encrypted for others.  I think a repair could be to use the HRMIS to lookup the email in LDAP, or some other creative solution.

Another solution would be to encrypt the mime messages once, for all recipients, and then draft one email, to all recipients.

sastels:
Fixes #1633 

# Description

Instead of having the email addresses in an environment variable, get them from the ldap query that we use to get the cert.

# Checklist:


davidheerema:
Once norm confirms he received the email as well I can approve my review

PR #1719, opened 2020-03-24
Validation dates

### LINKED ISSUE #1479: Validation for dates
## Summary

Date on when did it start page
Transaction date on money page

## Design Detail

Day = numbers only, 1-30, reformat 1 to 01
Month = numbers (or type month name?), 1-12, reformat 1 to 01
Year = numbers only, 1900-present

### LINKED ISSUE #1698: Move up to the beginning of page when I check the mandatory box
## Suggestion / question

With a new design, if the user forget to check the mandatory box (Privacy statement or What do you think could be affected?), the error message will be displayed. When the user checks the box, it will bring up to the beginning of the page instead of staying at it is as the old design. The user has to move the page down in order to click on <continue> button. Should we keep the old design? The user just needs to click on the check box and then <Continue> button 

## Use case

> What would this address?

amazingphilippe:
Fixes #1479 
Fixes #1698 

# Description

Adds validation on date fields

Fixes some issues with:
## ErrorSummary
- Pauses validation. This way, the focus does not jump around as validation is continuously being ran.
- Filters out bool errors. Bool errors are used when no error Message can be used (on day, month, year fields for example)

## FormArrayControls with errors
- Deals with one error at a time. This could be improved ?

## Next and Cancel buttons
- Resumes validation and submits form

# Required followup work

- [ ] Adding leading zeros to single date field digit
- [ ] Dealing with multiple errors on a single FormArrayControl
- [x] Conflicts with #1684. Either can be merged before or after
- [ ] Tweaking date validation conditions. They exist, but I haven't thought them through so much.

# Checklist:


amazingphilippe:
The new next and cancel buttons are failing form tests by calling submit twice. 
- We can't remove type=submit
- We can't remove the form.submit function call
- We don't want to call submit twice?

amazingphilippe:
>     * We don't want to call submit twice?

Call event.preventDefault()

f2/src/components/ErrorSummary/index.js
```diff
@@ -9,41 +9,54 @@ import { Li } from '../list-item'
...
+  useForm(props.onSubmit).pauseValidation()
+
   useEffect(() => {
     const summary = document
       .getElementById('error-summary')
       .getBoundingClientRect()
 
     window.scrollTo(0, summary.y - 16)
   })
+  console.log(errors)
```
amazingphilippe: remove this

f2/src/components/next-and-cancel-buttons/index.js
```diff
@@ -6,37 +6,60 @@ import PropTypes from 'prop-types'
...
-          textAlign="center"
-          w={{ base: '100%', md: 'auto' }}
-          ml={{ base: 0, md: 4 }}
-          mt={{ base: 4, md: 0 }}
-        >
-          <Trans id="button.cancelReport" />
-        </Button>
+export const NextAndCancelButtons = props => {
+  return (
+    <FormSpy subscription={{ hasValidationErrors: true }}>
```
amazingphilippe: Wrapping the buttons in a FormSpy this lets us have access to final-form's formAPI

f2/src/forms/HowDidItStartForm.js
```diff
@@ -13,9 +13,43 @@ import { TextArea } from '../components/text-area'
...
+  // if it has a value AND this value is a number below 12
+  if (
+    values.startMonth &&
+    (isNaN(values.startMonth) || values.startMonth > 12)
+  ) {
+    errors.whenDidItStart = 'whenDidItStart.startMonth.warning'
+    errors.startMonth = true
+  }
+  // if year contains 4 digits
+  if (values.startYear && values.startYear.length !== 4) {
```
kevinyang06: year can pass validation with non-number value

amazingphilippe: good catch!

PR #1728, opened 2020-03-25
remove submissionDate from the analyst email

### LINKED ISSUE #1701: remove submissionDate from the analyst email
## Summary

We added `submissionDate` to the report metadata to allow us to search the database for the number of reports submitted already (in case of a reboot). That field isn't needed in the analyst email - we should delete in `formatAnalystEmail()` right after we copy the data.

## Steps to reproduce

submit a report without `LDAP_UID` set and with an email address on the contact info screen. look at the bottom of the analyst report that you get.

kevinyang06:
Fixes #1701 

# Description

> We added submissionDate to the report metadata to allow us to search the database for the number of reports submitted already (in case of a reboot). That field isn't needed in the analyst email - we should delete in formatAnalystEmail() right after we copy the data.

# Any new packages installed?

> no

# Required followup work

> no
# Checklist:

### DIFF
--- a/f2/src/utils/formatAnalystEmail.js
+++ b/f2/src/utils/formatAnalystEmail.js
@@ -27,13 +27,13 @@ const formatReportInfo = data => {
         formatLineHtml('Report version:', data.appVersion) +
         formatLineHtml('Flagged:', selfHarmString),
     )
-
-  delete data.reportId // we delete the parts of the data object that we've displayed
-  delete data.submissionTime // so that at the end we can display the rest and ensure that
-  delete data.submissionDate //submissionDate does not needed in analyst report
-  delete data.language // we didn't miss anything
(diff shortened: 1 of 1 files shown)


f2/src/utils/formatAnalystEmail.js
```diff
@@ -30,6 +30,7 @@ const formatReportInfo = data => {
 
   delete data.reportId // we delete the parts of the data object that we've displayed
   delete data.submissionTime // so that at the end we can display the rest and ensure that
+  delete data.submissionDate //submissionDate does not needed in analyst report
   delete data.language // we didn't miss anything
```
sastels: The three lines of comment before were meant to be read as one long comment, so adding the new one in the middle makes things a bit confusing 😄  More confusing, perhaps, is why we did the three line comment like this in the first place. Could you move those three lines to before all these delete lines? Leaving in the line you added here just as it is.

PR #1732, opened 2020-03-26
only set submitted if submit succeeded

### LINKED ISSUE #1715: The error message " <!-- a padding to disable MSIE and Chrome friendly error page -->" should not be displayed
## Summary

If the report is not successful to submitted, the user goes back to the previous page and the following message is displayed

error

## Steps to reproduce

## Unresolved questions

sastels:
Fixes #1715

# Description

Only consider the report submitted if the submission is successful - this means that if the submission fails and the user navigates back to the confirmation page, they will not get the "thank you for submitting" message, but rather have the submit button that will allow them to try again.

To test a bad submission: run locally with `npm run dev` (not `node server.js`). Submissions will fail since you don't have a server running.

# Checklist:


f2/src/ConfirmationPage.js
```diff
@@ -112,7 +112,8 @@ const submitToServer = async (data, dispatch) => {
   console.log('Submitting data:', data)
   const response = await postData('/submit', data)
   const reportId = await response.text()
-  dispatch({ type: 'saveFormData', data: { reportId, submitted: true } })
+  const submitted = reportId && reportId.startsWith('NCFRS-')
```
amazingphilippe: Do we have to make sure it starts with "NCFRS-"? What if the report id prefix gets changed in the future? Could we instead check to see if report id is not an empty string?

sastels: If there is an error on the server we return the error message here instead of an id (kinda hacky... though if there's some sort of critical server error this might be all we get).

PR #1765, opened 2020-04-01
New env vars

davidheerema:
I have replaced the existing notify template var with these two new ones in the script.

TODO:
- Change the active environments to implement these changes (I can add them ahead of time and then remove the unused one later even)
- Steve to let me know the new template ID values


sastels:
This is changing a lot of files... why don’t I do what I should have done in the first place, and copy your pipeline changes directly into my branch.

davidheerema:
I tried to be clever and take master, make my changes to the script, then
merge that into your branch.  BUT, what I think happened is that the master
that I took was a few commits ahead of your branch and thats where these
extra changes are from!  I only wanted to take the one commit to the file...

Sorry If I made a mess here steve.

PR #1768, opened 2020-04-01
add prodVersion to report

### LINKED ISSUE #1703: Store prodVersion (1.1.0 currently) and include in analyst email
## Summary

in `App.js` we display the prod version as well as the gitsha. We don't store or send that data anywhere. We should store the data in the report, then save it to the database (which will happen automagically if it's in the report object) and include it in the analyst email.

## Steps to reproduce

Submit a report. "1.1.0" doesn't appear anwhere in the database or analyst email.

sastels:
Fixes #1703

# Description

Add the prod version (ie semantic version 1.1.0) to the report & email.

# Checklist:


f2/src/utils/formatAnalystEmail.js
```diff
@@ -18,13 +18,14 @@ const formatReportInfo = (data) => {
     selfHarmString = data.selfHarmWords
     returnString = `\n\n<h1>SELF HARM WORDS FOUND : ${selfHarmString}</h1>`
   }
+  const versionString = `${data.prodVersion} (${data.appVersion})`
```
amazingphilippe, line 21: Is this correct?: In the email, I get '1.1.0 (no version)' What is the difference between prod and app versions? Should we make it clearer?

sastels: 👍 Would probably be useful to do something like
```
1.1.0 (GitHub: 2a324f)
```
locally and on Heroku the GitHub hash doesn't work - I'll make that more expanitory too.

PR #1774, opened 2020-04-01
Contact info email validation not allowing valid email addresses

### LINKED ISSUE #1754: Contact info email validation not allowing valid email addresses
## Summary

Doesn't allow my CDS email address. I think it's because of the dash.

## Steps to reproduce

email address "sa@cds-snc.ca" not allowed. "sa@cdssnc.ca" is allowed.

## Unresolved questions

Are we missing something else? Is there a better regex on the web we should use?

kevinyang06:
Fixes #1754

# Description

> Doesn't allow my CDS email address. I think it's because of the dash.

# Any new packages installed?

> yes, email-addresses
https://www.npmjs.com/package/email-addresses

# Required followup work

> no
# Checklist:


f2/src/forms/ContactInfoForm.js
```diff
@@ -20,7 +20,7 @@ const validate = (values) => {
   //condition for an error to occur: append a lingui id to the list of error
   if (
     values.email !== '' &&
-    !new RegExp(/^\w+@[a-zA-Z_]+?\.[a-zA-Z]{2,3}$/).test(values.email)
+    !new RegExp(/^\w+@[a-zA-Z_-]+?\.[a-zA-Z]{2,3}$/).test(values.email)
```
sastels, line 23: I'm still concerned that we might mark legitimate emails as errors (for example, where the first part of the email contains a "+"). I'm thinking we should either relax it entirely and accept anything that has a "@" in it, or have a more complicated regex such as in https://www.w3resource.com/javascript/form/email-validation.php.

thoughts? @ngosset @davidheerema any thoughts?

ngosset:
I strongly suggest looking for a regularly maintained and popular
validation framework or e-mail & URL parsing library in npm and try using
that library. That's what I'd done in the Java world in the past - rely on
some apache java library that parses e-mails and urls in a try/catch block.
Update the library when new RFCs superseed the old ones and such.

What's a valid character in an e-mail and in URLs has changed over time
when new RFCs superseed old ones and such.

Validating e-mails is very complex:

https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression

davidheerema:
1000 times agree with Norm - need some tech specific for this purpose.

sastels:
lol @ `email-addresses`
>Want to see if something could be an email address? Want to grab the display name or just the address out of a string? Put your regexes down and use this parser!

PR #1799, opened 2020-04-08
Validate date

### LINKED ISSUE #1786: Add more validation on date
## Summary

1. If the user enters "0" on day, month or year (0-0-0000) --> No error is received.  It should give an error "Invalid date" or Start date/month/year must contain a real day/month/year"
2. If the user enters 30-02-2020 -> no error is received because no 30 on February. It should give an error "Invalid day or Date must contain a real day"
3. If the user enters 32-3-2020  or 2-13-2020-> the user will receive an error message "Start date must be in the past". It should be "Start date must contain a real day/month" or "Invalid day/month

## Steps to reproduce

## Unresolved questions

DianeLiu2019:
Fixes #1786

# Description
add more validation for date in howdiditstart page:  day in different month match the calendar . The day in Feb can't be >29 in leap year and can't be >28 in non-leap year, day , month , year can't be  '0'

# Checklist:

- [ x] I have added a comment to any confusing code


DianeLiu2019:
> This is a lot of validation! Great work! I tihnk we should add tests to make sure this works as we want it to work. We have some [validation testing on the phone number field](https://github.com/cds-snc/report-a-cybercrime/blob/5556d582caab57dbb7f7efec54dec524e9e78c2f/f2/src/forms/__tests__/ContactInfoForm.test.js#L24). You can do something similar!

f2/src/locales/fr.json
```diff
@@ -378,5 +378,5 @@
   "whenDidItStart.startMonth": "Mois",
   "whenDidItStart.startMonth.warning": "La date de l'incident doit comprendre un vrai mois",
   "whenDidItStart.startYear": "Année",
-  "whenDidItStart.startYear.warning": "L'année de la date de l'incident doit comprendre quatre chiffres"
+  "whenDidItStart.startYear.warning": "La date de l'incident doit comprendre un vrai année"
```
amazingphilippe, line 381: "La date de l'incident doit comprendre une vrai année" (add "e" in "une")

DianeLiu2019: fixed. Good catch

PR #1807, opened 2020-04-09
Datevalidtionmoneylostpage

### LINKED ISSUE #1800: More date validation on moneylost page

If the user enters "0" on day, month or year (0-0-0000) --> No error is received. It should give an error "Invalid date" or Start date/month/year must contain a real day/month/year"
If the user enters 30-02-2020 -> no error is received because no 30 on February. It should give an error "Invalid day or Date must contain a real day"
If the user enters 32-3-2020 or 2-13-2020-> the user will receive an error message "Start date must be in the past". It should be "Start date must contain a real day/month" or "Invalid day/month> One paragraph explanation of the bug.

## Steps to reproduce

## Unresolved questions

DianeLiu2019:
Fixes #1800 

# Description
add more validation for date in moneylost page: day in different month match the calendar . The day in Feb can't be >29 in leap year and can't be >28 in non-leap year.  Day , month , year can't be '0'
# Checklist:
- [ X] I have looked at my code on GitHub and it all looks good (ex: no random commented out code or console.logs)
- [ X] I have added and needed tests for my changes (in particular for new screens)
- [ X] I have added a comment to any confusing code


sastels:
> Do we want to introduce some sort of package to handle date validation? @sastels @ngosset

We could - this is a bit of a borderline case: validation is a bit complicated but not terrible (as email validation is, IMO). I don't think we should use a huge module like `moment` just for date validation, but maybe `date-fns` would work.

Alternatively we could just write our own code as we have in this PR and test it well. I would say in this case we should move the date validation code to the `src/utils` directory (and test to `src/utils/__tests__`. We can have tons of tests to ensure, for example, that every month must have the correct number of days, leap year works for the next 3 or 4 the year is not before 1980, etc.

ngosset:
Definitely not as bad as e-mail. We already have this PR so if someone wants to test it real well we could use it.

Alternatively I don't like reinventing the wheel.

- date-fns looks really simple. here's an example:

https://date-fns.org/v1.9.0/docs/isValid

Personally I'd use date-fns.
I would do a double prong check: 
Use regex to ensure user input is a specific format (yyyy-mm-dd)
Parse that into 3 integers, Use date-fns to make sure the date makes sense (No Feb 31st, etc)


> > Do we want to introduce some sort of package to handle date validation? @sastels @ngosset
> 
> We could - this is a bit of a borderline case: validation is a bit complicated but not terrible (as email validation is, IMO). I don't think we should use a huge module like `moment` just for date validation, but maybe `date-fns` would work.
> 
> Alternatively we could just write our own code as we have in this PR and test it well. I would say in this case we should move the date validation code to the `src/utils` directory (and test to `src/utils/__tests__`. We can have tons of tests to ensure, for example, that every month must have the correct number of days, leap year works for the next 3 or 4 the year is not before 1980, etc.

amazingphilippe:
> Use regex to ensure user input is a specific format (yyyy-mm-dd)
> Parse that into 3 integers, Use date-fns to make sure the date makes sense (No Feb 31st, etc)

We already have the individual fields. However, I think we should validate the whole date, and not just the individual fields.

amazingphilippe:
> https://date-fns.org/v1.9.0/docs/isValid

If we use v2, they also have isExist, which is good for avoiding converting Feb 31st into March 3. I think...

https://date-fns.org/v2.12.0/docs/isExists

davidheerema:
> If we use v2, they also have isExist, which is good for avoiding converting Feb 31st into March 3. I think...
> 
> https://date-fns.org/v2.12.0/docs/isExists

Phil, considering you already have the separate fields I would be surprised to learn of something more convenient than isExists!

amazingphilippe:
Closing this as work for date validation is overhauled in #1889

PR #1810, opened 2020-04-12
Refactor form default values

sastels:
# Description

Keep track of all the fields in the app and fail spectacularly if there are any mistakes.
(ie if there are fields that aren't in the `defaultValues.js` file then the corresponding form will not submit).

How this is done:
- there is a new file `defaultValues.js` that has all the fields in the app (together with a default value that is currently either an empty string or an empty list).
- each form now:
  - pulls from this file to get its initial values (previously they were hardcoded in the form file)
  - on submission calls the new function `areFieldsValid()` in `areFieldsValid.js` with the values in the form as well as the corresponding default values from `defaultValues.js`. If the form has keys that are NOT in the default file then the submission will not save the data to state nor navigate to the next page. The idea is to break the app so profoundly that it will be very difficult to not notice that we've made a mistake.
- the default values are also pulled in on the confirmation page (rather than having a second copy hardcoded there). This allows us to navigate to the confirmation page directly when running locally with `npm run dev`.

# Required followup work

- need to use `defaultValues.js` in the server to ignore invalid fields
- should also add to the storybook documentation around forms

# Checklist:


f2/src/ConfirmationSummary.js
```diff
@@ -17,53 +18,7 @@ import { WhatWasAffectedSummary } from './summary/WhatWasAffectedSummary'
...
-      suspectClues3: '',
-    },
-    evidence: {
-      files: [],
-      fileDescriptions: [],
-    },
-    location: { postalCode: '' },
-    contactInfo: { fullName: '', email: '', phone: '' },
-  },
+  formData: formDefaults,
```
amazingphilippe, line 21: 🥇

f2/src/utils/areFieldsValid.js
```diff
@@ -0,0 +1,14 @@
+const areFieldsValid = (data, defaults) => {
+  let valid = true
+  Object.keys(data).forEach((field) => {
+    if (Object.keys(defaults).indexOf(field) === -1) {
+      valid = false
+      console.log(
+        `ERROR: field ${field} not in defaults ${Object.keys(defaults)}`,
```
amazingphilippe, line 7: spectacular!

sastels:
> Looks amazing. Do you think there is a way to bring these default values to format the analyst email more automagically?

not easily I don't think. The thing about the analyst email is that it takes fields from different screens and puts them together into sections in the email. And does things like replace "other" options with the "other" text fields. I think that trying to make that code more compact would make it easier to introduce errors when changing fields around in the future :/

PR #1816, opened 2020-04-14
phase banner tag more obvious

amazingphilippe:
# Description

Tweak to the beta banner tag

# Checklist:

### DIFF
--- a/f2/src/components/phase-banner/index.js
+++ b/f2/src/components/phase-banner/index.js
@@ -5,22 +5,21 @@ import { Flex, Tag } from '@chakra-ui/core'
 import { Text } from '../text'
 import { Layout } from '../layout'
 
-export const PhaseBanner = props => (
+export const PhaseBanner = (props) => (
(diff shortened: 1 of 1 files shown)


f2/src/components/phase-banner/index.js
```diff
@@ -5,22 +5,21 @@ import { Flex, Tag } from '@chakra-ui/core'
...
 import { Layout } from '../layout'
 
-export const PhaseBanner = props => (
+export const PhaseBanner = (props) => (
   <Flex bg="gray.100">
     <Layout>
       <Flex align="center" py={4}>
         <Tag
           fontFamily="body"
+          fontWeight="bold"
```
sastels, line 14: do we need both family and weight?

amazingphilippe: Yes! Family set the font family, as in "Noto Sans", we don't access the font family as "Noto Sans Bold" So weight is necessary!

PR #1820, opened 2020-04-15
Anonymous flow opt-in

kevinyang06:
Fixes #1804

# Description

> Changing the way we ask for anonymous reporting:

Replace the yes/no options by a single checkbox
Remove validation, this is an optional field
If the box is checked, display a warning below the checkbox to explain to users why we don't encourage anonymous reports.
# Any new packages installed?

> no

# Required followup work

> no

# Checklist:

- [ ] I have looked at my code on GitHub and it all looks good (ex: no random commented out code or console.logs)
- [ ] I have added and needed tests for my changes (in particular for new screens)
- [ ] I have added a comment to any confusing code
- [ ] I have added documentation to any modified front-end code. (Or added missing documentation)

PR #1834, opened 2020-04-16
Anonymous summary

### LINKED ISSUE #1771: Anonymous reporting on the summary page.
## Suggestion / question

On the summary, we could add a block to see whether the user chose to be anonymous or not. Also give the oppportunity to change this. We could put a warning banner if the user chose to report anonymously.

## Use case

Giving a last opportunity to opt in reporting with identification (not anonymously)

Here's how it could look:

Location anonymous

amazingphilippe:
Fixes #1771 

# Description

Adds a summary block for anonymous opt-in on summary page. Effectively replaces the contact info bock, as they are mutually exclusive.

# Required followup work

Merge #1820 before this.

# Checklist:


amazingphilippe:
> LGTM. Going for anonymous to non-anonymous is a one way trip (ie then there's no link back from the confirmation page) but I assume that's what we want?

Clearly I was not thinking straight because this is weird, changing your mind doesn't mean we will re-capture the postal code, and we just keep the city and province. It feels a bit mind bending at the moment, lets hold it off a bit more, Maybe there's a better way to switch between anonymous and non-anonymous

PR #1835, opened 2020-04-16
number of submissions and availability not rolling over with new day

### LINKED ISSUE #1821: number of submissions and availability not rolling over with new day
## Summary

- when the day rolls over, we are not resetting the number of submissions nor letting CAFC redirect to us
- in `saveRecord.js` the `currentDate` is set on program run. We should move it into the `getReportCount()` function.

- in `checkIfAvailable.js` we are using the value of `availableData.numberOfSubmissions`, even if a new day has started, so that after day rollover we keep telling CAFC that we are not available. We should add a check that returns `true` if the current date is different from the date in `availableData.lastRequested` (note that we probably don't want to do a database count here since this will be hit every time someone goes to the CAFC site)

kevinyang06:
Fixes #1821

# Description

> when the day rolls over, we are not resetting the number of submissions nor letting CAFC redirect to us

# Any new packages installed?

> no

# Required followup work

> no

# Checklist:


davidheerema:
@kevinyang06, there is a lot of merge conflicts and additional files in here..

Also, a hotfix should get merged against rcmp-release-prod branch

PR #1838, opened 2020-04-17
number of submissions and availability not rolling over with new day

### LINKED ISSUE #1821: number of submissions and availability not rolling over with new day

kevinyang06:
Fixes #1821

# Description

> when the day rolls over, we are not resetting the number of submissions nor letting CAFC redirect to us

# Any new packages installed?

> no

# Required followup work

> no
# Checklist:

### DIFF
--- a/f2/src/utils/checkIfAvailable.js
+++ b/f2/src/utils/checkIfAvailable.js
@@ -10,7 +10,36 @@ else console.info('Availability configured')
 // availableData is { numberOfSubmissions, numberOfRequests, lastRequested }
 const isAvailable = (availableData) => {
   try {
-    const currentTime = new Date()
+    const currentTime = new date()
+    //convert current date and lastRequested date with format of 00/00/0000
+    const currentDate =
+      (new Date().getDate() > 9
+        ? new Date().getDate()
+        : '0' + new Date().getDate()) +
+      '/' +
+      (new Date().getMonth() > 8
+        ? new Date().getMonth() + 1
+        : '0' + (new Date().getMonth() + 1)) +
+      '/' +
+      new Date().getFullYear()
+    console.log('current date is: ' + currentDate)
+
+    const lastRequestedDate =
+      (lastRequested.getDate() > 9
+        ? lastRequested.getDate()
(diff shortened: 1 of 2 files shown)


f2/src/utils/checkIfAvailable.js
```diff
@@ -11,6 +11,33 @@ else console.info('Availability configured')
...
+      (availableData.lastRequested.getMonth() > 8
+        ? availableData.lastRequested.getMonth() + 1
+        : '0' + (availableData.lastRequested.getMonth() + 1)) +
+      '/' +
+      availableData.lastRequested.getFullYear()
+
+    if (currentDate !== lastRequestedDate) {
+      availableData.numberOfSubmissions = 0
+      return true
+    }
```
sastels, line 40: ok i couldn't figure the "suggestion" out for this much code. But I think we can replace this whole new block with
```
const lastRequested = availableData.lastRequested
if (!lastRequested || (currentTime.getDate() !== lastRequested.getDate()) || (currentTime.getMonth() !== lastRequested.getMonth()) {
    availableData.numberOfSubmissions = 0
    return true
}
```

f2/src/utils/saveRecord.js
```diff
@@ -52,6 +45,13 @@ async function saveRecord(data, res) {
   }
 }
 async function getReportCount() {
+  const date = new Date()
+  const currentDate =
```
obrien-j, line 49: Just threw up a comment into slack, but this is the better place.

Check out the image below as a simplified version of the code below.

image

PR #1843, opened 2020-04-17
update changelog to v1.1.0 prod release

### LINKED ISSUE #1818: update Changelog for v1.1.0
## Summary

- update `CHANGELOG.md` to reflect the changes that made it into the prod version released today.
- also increment the version to 1.2.0

## Motivation

Keep track of what's in prod

sastels:
fixes #1818 

# Description

Update the changelog for the 1.1.0 prod release

# Checklist:


f2/CHANGELOG.md
```diff
@@ -1,24 +1,52 @@
 # Changelog for report a cybercrime product
 
-## V1.1.0 (staging: 2020-03-16)
+## V1.1.0 (2020-04-14)
```
amazingphilippe, line 3: V1.2.0 ?

sastels: We never actually got the previous 1.1.0 staging out the door (ie didn't put it to prod) so I figured our current release was 1.1.0 instead.

amazingphilippe: So should we stick to 1.1.0 in initialState ?

sastels: ah, I see your point! sorry, slow this morning. So the 1.2.0 in initialState is what the code is going forward. ie master from now on will be 1.2.0 rather than 1.1.0. I though doing it this way would make it easier to do releases, ie wouldn't have to change the counter. but... it's confusing. And what if the next release was "1.1.1" or "2.0.0". Alright, I'm convinced, let me change the initialState version to something else. Possibly something like "1.1.0-dev" or something. Then we change it to 1.2.0 or whatever when we actually cut the next release.

amazingphilippe: I see, makes sense. W'ere working on 1.2.0, but we are releasing the changelog from 1.1.0. Makes sense to me either way

PR #1854, opened 2020-04-21
Staging 1.2.0

sastels:
# Description

release candidate 1.2.0


f2/src/AnonymousPage.js
```diff
@@ -27,14 +28,14 @@ export const AnonymousPage = () => {
               <Lead>
                 <Trans id="anonymousPage.intro" />
               </Lead>
-
+              {console.log(doneForms)}
```
khalidelaggan, line 31: Was this added for testing? should we remove it?

amazingphilippe: Thought i removed it! Can we just remove it before merging?

sastels: whoa, good catch @khalidelaggan ! Removed.

davidheerema: was it removed from master as well?

sastels: Nope, just this PR

PR #1862, opened 2020-04-23
exclude .png file been uploaded

### LINKED ISSUE #1857: CAFC does not accept .png files
Users are uploading .png files via the NCFRS. The CAFC rejects all of these file times.

The work around is for the Analyst to screen capture the .png image, save as a .pdf and upload to the CAFC.

.png files were not included in the original list of file types accepted by the CAFC so suggestion that these should be excluded from our app.

DianeLiu2019:
Fixes #1857

# Description
as suggested by RCMP Stephanie Gauthier: 
.png files were not included in the original list of file types accepted by the CAFC so suggestion that these should be excluded from our app.
This PR is to forbid .png file to be uploaded and add more permitted file type in the String to inform victim

# Checklist:

- [ x] I have added a comment to any confusing code

### DIFF
--- a/f2/src/locales/en.json
+++ b/f2/src/locales/en.json
@@ -152,7 +152,7 @@
   "evidencePage.details": "If you can, include things such as:",
   "evidencePage.fileDescription": "Describe what this file shows",
   "evidencePage.fileSize": "You can add a maximum of 3 files, each up to 4 MB in size each.",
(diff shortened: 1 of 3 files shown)


DianeLiu2019:
Team decide to deal with the issue in other way. For example, have our application accept PNG and then do an image conversion to JPEG before emailing the report. decision need to de decide.

PR #1863, opened 2020-04-23
Move submitted and reportId out of formData

sastels:
# Description

Fixes a bug where second submission (ie trying after a first failure) would always be rejected.

To test locally:

- `npm run prod` and go through app. Submission works.
- Go back with browser back button. Confirmation page knows you've already submitted and has your report id at the bottom.
- Now go forward to the thankyou page and "create a new report". Verify that this still works and you can submit a new report.
- Next go back to the landing page, and stop `server.js` in your terminal. Fill out the report. The submission should fail (since the server isn't running).
- Click "Back to your report" - the confirmation page should have Edit links and a submit button.
- Run `node server.js` in your terminal and wait a few seconds for it to boot.
- Click "Submit report". This time the submission should succeed.

# Checklist:

### DIFF
--- a/f2/src/FinalFeedbackPage.js
+++ b/f2/src/FinalFeedbackPage.js
@@ -27,7 +27,7 @@ async function postData(url = '', data = {}) {
   return await response
 }
 
-const submitToServer = async data => {
+const submitToServer = async (data) => {
   console.log('Submitting finalFeedback:', data)
   await postData('/submitFeedback', data)
 }
@@ -45,10 +45,10 @@ export const FinalFeedbackPage = () => {
               <Trans id="finalFeedback.title" />
             </H1>
             <FinalFeedbackForm
-              onSubmit={data => {
+              onSubmit={(data) => {
                 submitToServer(data)
                 setState((state.doneFinalFeedback = true))
-                if (state.formData && state.formData.submitted)
+                if (state.formData && state.submitted)
                   history.push('/thankYouPage')
                 else history.push('/finalfeedbackthanks')
               }}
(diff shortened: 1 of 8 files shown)


f2/src/FinalFeedbackPage.js
```diff
@@ -45,11 +45,10 @@ export const FinalFeedbackPage = () => {
               <Trans id="finalFeedback.title" />
             </H1>
             <FinalFeedbackForm
-              onSubmit={data => {
+              onSubmit={(data) => {
                 submitToServer(data)
                 setState((state.doneFinalFeedback = true))
-                if (state.formData && state.formData.submitted)
```
khalidelaggan, line 51: Is this where the default values were causing the issue?

f2/src/components/EditButton/index.js
```diff
@@ -7,9 +7,9 @@ import { useStateValue } from '../../utils/state'
 
 export const EditButton = ({ path, label }) => {
   const { i18n } = useLingui()
-  const [{ formData }] = useStateValue()
+  const [{ submitted }] = useStateValue()
 
-  return formData.submitted ? null : (
+  return submitted ? null : (
```
khalidelaggan, line 12: Just a quick question, what is the difference between formData and submitted again?

DianeLiu2019:
I tested it , it works well.

f2/src/FinalFeedbackPage.js
sastels: the issue was that we were saving `submitted` in `state.formData`, and then the server would reject the submission as invalid (for having extra fields). The change here reflects that we have now moved `submitted` to `state`

f2/src/components/EditButton/index.js
sastels: `formData` holds the forms the user has filled out. `submitted` is an extra boolean variable that the app uses to know whether the user has already submitted or not (so, for example, if they navigate back from the thank you page the app doesn't let them submit the same report again)

Issue #1842, opened 2020-04-17
The same fix on #1800 and #1808 should be applied on Transaction Date


lmcbhvu:
## Summary

The same fix on #1800 and #1808 should be applied on Transaction Date in "How were your money or finances affected?" page

## Steps to reproduce

> How exactly can the bug be reproduced? Be very specific.

## Unresolved questions

> Are there any related issues you consider out of scope for this issue that could be addressed in the future?

DianeLiu2019:
fixed the issue in PR#1889

PR #1889, opened 2020-04-29
Date validation util

### LINKED ISSUE #1800: More date validation on moneylost page

amazingphilippe:
Fixes #1800 #1842

# Description

- Added a validation utility for dates
- Tests for dates are moved to the util
- Updated ErrorSummary and FormArrayControl to be able to display multiple errors for one field.
- Updated forms

# New packages:
- [date-fns](https://date-fns.org/docs/Getting-Started)

# Follow up work:

- We could replace date validation messages with default translation, instead of having distinct translations for separate pages. The translations should be identical for both.

# Checklist:


f2/src/utils/validateDate.js
```diff
@@ -0,0 +1,56 @@
...
+  if (day > getDaysInMonth(new Date(year, monthIndex)) || day < 1)
+    isDay = 'notDay'
+
+  // month is between 0 and 11
+  if (monthIndex > 11 || monthIndex < 0) isMonth = 'notMonth'
+
+  // is not in the future, while all fields are valid
+  !isMonth && !isDay && !isYear && isFuture(date) && validate.push('isFuture')
+
+  //Year is over 1867 :eyes:
```
sastels, line 35: this is a requirement?

f2/src/components/ErrorSummary/ErrorSummary.stories.mdx
```diff
@@ -19,9 +19,18 @@ import { Trans } from '@lingui/react'
 
 The ErrorSummary component uses final-form validation to assign errors to fields. This can be set in a `validate` function that returns an error object. The object is a list of field names as keys, pointing to lingui ids as values. This function is usually defined before exporting the form component.
 
+When you can have multiple errors on a single field, you can also pass an array to the errors object.
+
```
sastels, line 23: Why are we changing these components?

f2/src/utils/__tests__/validateDate.test.js
```diff
@@ -0,0 +1,137 @@
...
+  it('fails incorrect transaction date in non-Leap year', () => {
+    expected = ['notDay']
+    expect(validateDate('2017', '2', '29')).toEqual(
+      expect.arrayContaining(expected),
+    )
+    expect(validateDate('2019', '02', '29')).toEqual(
+      expect.arrayContaining(expected),
+    )
+  })
+})
```
sastels, line 137: awesome testing!

f2/src/utils/validateDate.js
amazingphilippe: No, I just don't know why we are accepting dates as far a year 1000, or even year 0001. I don't think this would affect people really, just a troll-barrier-easter-egg. We can remove this condition if we want!

f2/src/components/ErrorSummary/ErrorSummary.stories.mdx
amazingphilippe: To accept multiple errors on field sets. Especially the date field set, where we can have multiple distinct errors.

f2/src/utils/validateDate.js
DianeLiu2019: Actually I feel it is good idea to set  the year 150 years from now. If report input the year before 1867. It probably they made mistake. who can live that long and there was no computer at that time. I also check UK'S antifraud web: https://reporting.actionfraud.police.uk/reporting-form  The effective year is after 1900. That's my opinion. I will post a message  to get other team members opinion

amazingphilippe:
> LGTM! only thought I had was we have `whenDidItStart.error.notDay`, `transactionDate.error.notDay`, etc, maybe we want to change them to just `dateValidation.error.notDay`, etc if they're going to always have the same text. Just an idea.

Yes, Added to follow up work. We have that default.x translation prefix we could use

PR #1891, opened 2020-04-30
Stop accepting zero bytes attachments.

sdzhangtao:
Fixes #(1888)
When upload files, user could upload a zero bytes file, which serves no value
# Description
When user uploads zero bytes files, it could cause error for content moderator etc
When user selects the file, app will check the file size, if it's zero, will use a local pop up to display the error message
# Any new packages installed?
No
> Give details

# Required followup work
Not sure the Error message is up to the standard
The translation for the French text

# Checklist:


sdzhangtao:
Changed == to ===
Will need to learn the PR process

PR #1902, opened 2020-05-04
required contact fields

### LINKED ISSUE #1873: The user get stuck at contact info page if the user wants remove to phone or email address
## Summary

After choose to submit the non-anonymous report with entering name, gmail and phone. AT summary page, the user wants to edit contact info and remove email or phone. After the user remove the phone number,  the error message "enter a phone number that contains 10 digits" keep display when the user press continue button. Similarly  it happened with email address. 

## Steps to reproduce

- Choose  to submit non-anonymous report
- Enter name, email and phone at contact detail info page
- At summary, edit that section and try to remove phone or email and press continue

## Unresolved questions

amazingphilippe:
Fixes #1873 

# Description

Forces the user to include contact information in non-anonymous report. Removes the skippable contact info.

# Required followup work

Review the quality of reports with these new changes. 

# Checklist:


DianeLiu2019:
I tested it . It seems no error message shown when the full name field is empty. is that what you want?

amazingphilippe:
> full name can still be blank if user go back from summary page and remove their names

I think I fixed it. We should also check with any other validation we have. In this case, when returning from summary page, deleting the name also deleted the key. So I had to check if the key existed in the values when I validated, THEN if they had a non-empty value : `const email = 'email' in values && containsData(values.email)`

PR #1908, opened 2020-05-06
Enforce consent better

### LINKED ISSUE #1602: Enforce consent better
## Summary

fix consent-related edge cases

## Design Detail

There are edge cases where someone could get into a state where they are navigating through the app but their consent has not been recorded (ex: cancel a report and then use back button go go back to the report flow). If this happens the app should redirect the user back to the consent page.

kevinyang06:
Fixes #1602

# Description

> There are edge cases where someone could get into a state where they are navigating through the app but their consent has not been recorded (ex: cancel a report and then use back button go go back to the report flow). If this happens the app should redirect the user back to the consent page.

# Any new packages installed?

> no

# Required followup work

> anonymous page need to fix before apply this enforce to consent page

# Checklist:


sdzhangtao:
I don't have much experience doing React page flows all that. Just feel something is unusual here. Feels like there must be another way of enforcing the page flow logic, a centralized place, instead of duplicating the check into each page. I have known HoC. Or this enforcement is only required during the submit time. 
Can we only enforce that during the submit time?

sdzhangtao:
This enforcing redirect routing logic sounds like a cross cutting concern. It's the best to keep in one place for better maintainability. Imagine if there is another logic such as this. 
We could build a custom Route component that encapsulate such enforcing logic. It's similar idea to enforcing authentication redirect. I found a sample in this link:
https://reacttraining.com/react-router/web/example/auth-workflow

kevinyang06:
> This enforcing redirect routing logic sounds like a cross cutting concern. It's the best to keep in one place for better maintainability. Imagine if there is another logic such as this. It's far easier to have the logic in one place.
> We could build a custom Route component that encapsulate such enforcing logic. It's similar idea to enforcing authentication redirect. I found a sample in this link:
> https://reacttraining.com/react-router/web/example/auth-workflow

can we make this merge first then I will start working on your suggestion, this way we can have at least something works in the master and prod.

amazingphilippe:
> The code works, but it's better to refactor the code for maintainability. See my comments

We should create an Issue explaining the idea! This way it doesn't get lost and anyone can fix the issue when they have the time. Btw, I love the idea of managing flows in one place. We never really had the chance to think about this really.

PR #1931, opened 2020-05-20
Added ability to submit from json file

davidheerema:
# Description

> I wrote some sketchy code to allow this utility to submit data from a .json file as well.  That way we can take a successful record from the database as test-data and modify it easily.

# Any new packages installed?

> Nope

# Required followup work

# Checklist:


f2/src/utils/submitReportToServer.js
```diff
@@ -33,8 +39,23 @@ async function submitReportToServer(url = '', data = {}) {
...
-let data = JSON.parse(JSON.stringify(formDefaults))
-data.contactInfo.fullName = 'Mallory'
-// data.contactInfo.extra = 'extra field!'
+const path = './TestData.json'
 
-submitReportToServer('http://localhost:3000/submit', data)
+fs.access(path, fs.F_OK, (err) => {
+  if (err) {
+    console.error(err)
+    console.info('Using data from JSON Object')
```
sdzhangtao, line 47: What's the difference between submit from JSON object and submit from JSON file?

davidheerema: This file was originally a little test utility made by steve.  Ignore all the Mallory stuff if you like.

My idea is a) if there is a Test.json file present, use that file and the data inside it to submit to the server.

If that file is not present, load from the defaultvalues.js

f2/src/utils/submitReportToServer.js
```diff
@@ -33,8 +39,23 @@ async function submitReportToServer(url = '', data = {}) {
...
+    // data.contactInfo.extra = 'extra field!'
+    submitReportToServer('https://report-dev.con.rcmp-grc.gc.ca/submit', data)
+    return
+  }
+  console.info('Using data from JSON File')
+  let rawdata = fs.readFileSync('Test.json');
+  let data = JSON.parse(rawdata);
+  data.contactInfo.fullName = 'MalDlory'
+  // data.contactInfo.extra = 'extra field!'
+  submitReportToServer('https://report-dev.con.rcmp-grc.gc.ca/submit', data)
```
sdzhangtao, line 59: The only thing I can think of is to add this to the fetch call to give out more error output other then just status code 500 or 422.
catch(function (error) {
    console.log({ error })
  })

PR #1966, opened 2020-06-01
Remove is checked

### LINKED ISSUE #1626: isChecked on checkboxAdapter component does nothing
## Summary

We set isChecked based on what is in the local state data. This is pointless. The field will do this check automatically. We should remove these properties from forms

## Steps to reproduce

1. Remove `isChecked` from all `<CheckboxAdapter>`
2. See that it has no effect when going from page to page.

## Unresolved questions

Do the change wherever it needs to be done

justinr86:
Fixes #1626 

# Description

Removed isChecked prop from checkbox and radio button adapters.

# Any new packages installed?

No new packages.

# Required followup work

No further work.

# Checklist:


DianeLiu2019:
do we need also delete the "Ischecked" form f2/src/components/radio/index.js and f2/src/components/checkbox/index.js?

DianeLiu2019:
> > do we need also delete the "Ischecked" form f2/src/components/radio/index.js and f2/src/components/checkbox/index.js?
> 
> The **isChecked** property on the component is used to display conditional content. For example, the **How Did It Start** page has checkboxes that will display a textarea when selected. This property needs to remain in the component as it still has functionality. Setting it in the form is not necessary as the state is already managed.

Thank you for explaining to me.

PR #1975, opened 2020-06-03
Encrypted E-mail and attachments redesign

### LINKED ISSUE #1704: The attachment symbol is always displayed even no attachment in analyst email
## Summary

The attachment symbol is always displayed even no attachment in analyst email

## Steps to reproduce

fileAttached

## Unresolved questions

### LINKED ISSUE #1900: E-mail fails to send to analyst if larger than 10MB
## Summary

We are currently using CDS Notify's SMTP server to send the report to the analyst. This service has a 10MB file limit per e-mail (after email is base64 encoded).

A report with 3 large images (up to 4MB each) puts us over 12MB so we get an error:

(node:1) UnhandledPromiseRejectionWarning: Error: Message failed: 552 5.3.4 Message is too long.                           

## Steps to reproduce

Create a report with 3 large files

## Unresolved questions

We need to determine a strategy to fix this and then implement. Possible options:
1) Send files in their own e-mails
2) Use "Shared Access Signature" and give the analyst a Link to the file in Azure Blob Storage
3) Use a different SMTP server for reports?

### LINKED ISSUE #1953: File name was too long, analyst could not open it (report received on May 26)
## Summary

The file that was attached had too long of a name and analysts were not able to open the report until is was downloaded to the database. 

## Steps to reproduce

Submit a file with a very long file name.

ngosset:
Fixes #1953 
Fixes #1900 
Fixes #1704 (All e-mails will have a minimum of .html report attachment now)

This PR changes how we send file attachments and also re-designs how we send the report to fix some usability issues with groupwise.

The way we built the HTML report before wasn't entirely correct, but Outlook was more forgiving than groupwise. 

**Before**: Analysts were sent an encrypted html e-mail with all files attached and an html body. But groupwise did not render the HTML, instead the e-mail showed up as an ".txt" attachment with tedious process to view, and files over 10MB would cause e-mail to fail

**Now**: Analyst are sent a plain encrypted e-mail with a single report file ".htm" attached. Analyst can open this html in browser of their choosing. HTML report has download links embedded for additional attachments.

To test:

Add the following to your .env file

```
# How many days for a read-only token for blob storage to expire
BLOB_SAS_DAYS_EXPIRY=5

# Restrict Azure blob downloads to this IP range 
BLOB_SAS_IP_LOWER=000.000.000.000
BLOB_SAS_IP_UPPER=255.255.255.255
```

Also update the .env to configure your instance to send you encrypted test reports at your rcmp e-mail address.
```
LDAP_UID=(your hrmis)
MAIL_TO=(your e-mail)
```

Ii you don't have a HRMIS and RCMP e-mail, you can test by leaving the above ENV vars blank. Instead set the "MAIL_LOCAL" env var to an e-mail address of your choosing. This will send you an unencrypted e-mail with unencrypted download links.

Once configured and running:
Fill out a report as normal. Attach files
You will get an encrypted e-mail with a .htm attached. This .htm is the report. Within the report will be download links for any attachment files.

Other considerations:
I'm still a modern JS novice. Some of this could probably be done more async for performance?
This encrypts files before we save them to Azure Blob Storage. So this means we won't be able to recover these files in the future unless we have an analyst decrypt them for us?


f2/server.js
```diff
@@ -143,15 +143,19 @@ const allowedReferrers = [
...
-// These can all be done async to avoid holding up the nodejs process?
-async function save(data, res) {
+// Moved these out of save() and to their own function so we can block on 'saveBlob' to get the SAS link
+// without holding up the rest of the 'save' function
+async function saveBlobAndEmailReport(data) {
   var converted = await convertImages(data.evidence.files)
   data.evidence.files.push(...converted.filter((file) => file !== null))
-  saveBlob(data)
-
+  // Await on this because saveBlob generates the SAS link for each file
```
khalidelaggan, line 151: +1

PR #1999, opened 2020-06-04
Elaborate Availability Logging & Remove duplicate logic

davidheerema:
# Description

Availability logic moved from server.js to checkIfAvailable.js, this will make things easier to maintain as all relevant logic is in one place.


f2/src/utils/checkIfAvailable.js
```diff
@@ -3,10 +3,9 @@ require('dotenv').config()
 const submissionsPerDay = process.env.SUBMISSIONS_PER_DAY
 const secondsBetweenRequests = process.env.SECONDS_BETWEEN_REQUESTS
 
-if (!submissionsPerDay || !secondsBetweenRequests)
-  console.error('ERROR: Availability not configured')
-else console.info('Availability configured')
-
+if (!submissionsPerDay || !secondsBetweenRequests) {
+  console.error('ERROR: SUBMISSIONS_PER_DAY or SECONDS_BETWEEN_REQUESTS not configured.  The Server will constantly report unavailable as a result.')
```
justinr86, line 7: If **availableData.lastRequested** is not set **OR** if there has not been a request today the service will return **true**. Only if these checks evaluate **false** will the service check **submissionsPerDay** or **secondsBetweenRequests**. We should change the message or rearrange the logic to be consistent.

justinr86:
Availability logic moved from server.js to checkIfAvailable.js, this will make things easier to maintain as all relevant logic is in one place.

ngosset:
@Justin Richardson Everything (CHECK_REFERER, Submission per day, time between sub, /available /stats) seems to work except the TOTP bypass

Trying to use a TOTP code also leads to:
Error: Can't set headers after they are sent.
    at SendStream.headersAlreadySent (/Users/normangosset/projects/report-a-cybercrime/f2/node_modules/send/index.js:390:13)
    at SendStream.send (/Users/normangosset/projects/report-a-cybercrime/f2/node_modules/send/index.js:617:10)
    at /Users/normangosset/projects/report-a-cybercrime/f2/node_modules/send/index.js:774:12
    at FSReqCallback.oncomplete (fs.js:172:5)

PR #2045, opened 2020-06-18
remove The City and Province base on the postal code when report anonymous

### LINKED ISSUE #2038: Should City and Province base on PC be removed if it is anonymous report?
## Summary

The City and Province base on the postal code should be removed in the analyst email if the report is an anonymous. It does not make sense to have it because there is no postal code in the report and city/province is already included in the report

file1 (1)

## Steps to reproduce

## Unresolved questions

DianeLiu2019:
Fixes #2038(issue)

# Description
remove "The City and Province base on the postal code " in the analyst email if the report is an anonymous. also remove " postalCity = 'Location lookup failed'
    and postalProv = 'Location lookup failed' to avoid showing  "Location lookup failed" when report anonymous. 

# Checklist:


f2/src/utils/formatAnalystEmail.js
```diff
@@ -66,7 +66,7 @@ const formatVictimDetails = (data) => {
   let postalProv = ''
   try {
     let location = zipcodes.lookup(data.location.postalCode)
-    if (location === undefined) {
+    if (data.location.postalCode && location === undefined) {
```
kevinyang06, line 69: this cause error in console 'TypeError: Cannot read property 'city' of undefined' 
now with the new condition, the logic try to fetch location.city when it is anonymous report.

PR #2065, opened 2020-06-24
Attach stemmer based on language

### LINKED ISSUE #2063: Key word refinement (#2041) does not work well
## Summary

Key word refinement (#2041) does not work well. Yes, there is a  limitations of the stemmer but it does not catch well even with the words which have an extra "s", "d" or "ed" at the end
1. killed, killing instead of kill (but it does for kills)
2. suicide instead of suicide
3. overdosed instead of overdose
4. it is over instead of it's over
.....

## Steps to reproduce

## Unresolved questions

justinr86:
Fixes #2063 

# Description

PR #2041 adds a stemmer to identify root words. This update will use the English or French stemmer depending on language of submission. 

# Any new packages installed?

N/A

# Required followup work

N/A

# Checklist:

### DIFF
--- a/f2/src/utils/selfHarmWordsScan.js
+++ b/f2/src/utils/selfHarmWordsScan.js
@@ -9,15 +9,19 @@ const selfHarmWords = selfHarmString
   .map((w) => unidecode(w.trim().toLowerCase()))
 console.info(`Self harm word list: ${selfHarmWords}`)
 
-//Patches stem() and tokenizeAndStem() to String as a shortcut to PorterStemmer.stem(token)
-natural.PorterStemmer.attach()
-natural.PorterStemmerFr.attach()
-
 //Scan form data for self harm key words.
 const selfHarmWordsScan = (data) => {
   try {
     const jsonString = unidecode(JSON.stringify(data).toLowerCase())
     const json = JSON.parse(jsonString)
+
+    //Patches stem() and tokenizeAndStem() to String as a shortcut to PorterStemmer.stem(token)
+    if (json.language === 'fr') {
+      natural.PorterStemmerFr.attach()
+    } else {
+      natural.PorterStemmer.attach()
+    }
+
     let keyWordMatch = []
 
     for (let key in json) {


DianeLiu2019:
not sure if this PR fix the issue
I tested:  "killed , killing and kill "all work well. 
 'suicide' works, but 'suicides, suicided' don't work. 
'overdose' works , but 'overdosed 'doesn't work
 'it's over' works, but 'it's over' doesn't work .

justinr86:
> not sure if this PR fix the issue
> I tested: "killed , killing and kill "all work well.
> 'suicide' works, but 'suicides, suicided' don't work.
> 'overdose' works , but 'overdosed 'doesn't work
> 'it's over' works, but 'it's over' doesn't work .

This goal of this update is to add a [stemmer](https://en.wikipedia.org/wiki/Stemming) to the self harm key word scan. This stemmer has limitations. Use this [website](http://9ol.es/porter_js_demo.html) to see what the stemmer will produce. For example, the suffix **ed** is removed in an attempt to find the root word, so **suicided** becomes **suicid**. Both the original word and the stemmed version are scanned in an attempt to find any matches.

The important thing is that there are no false positives and exact matches are flagged. This PR will expand the words that match but it isn't perfect.

DianeLiu2019:
> > not sure if this PR fix the issue
> > I tested: "killed , killing and kill "all work well.
> > 'suicide' works, but 'suicides, suicided' don't work.
> > 'overdose' works , but 'overdosed 'doesn't work
> > 'it's over' works, but 'it's over' doesn't work .
> 
> This goal of this update is to add a [stemmer](https://en.wikipedia.org/wiki/Stemming) to the self harm key word scan. This stemmer has limitations. Use this [website](http://9ol.es/porter_js_demo.html) to see what the stemmer will produce. For example, the suffix **ed** is removed in an attempt to find the root word, so **suicided** becomes **suicid**. Both the original word and the stemmed version are scanned in an attempt to find any matches.
> 
> The important thing is that there are no false positives and exact matches are flagged. This PR will expand the words that match but it isn't perfect.

Thank you for explaining it.

PR #2067, opened 2020-06-26
Limit content moderator

### LINKED ISSUE #1578: Rate limit content moderator calls
We need to rate limit how often calls get made to content moderator.  The code should be written to only pass one call per second.

When this issue is fixed notify dave so he can configure azure to stop paying $2

justinr86:
Fixes #1578  

# Description

Use asynchronous queue to call content moderator once per second.

# Any new packages installed?

N/A

# Required followup work

N/A

# Checklist:

### DIFF
--- a/f2/src/utils/scanFiles.js
+++ b/f2/src/utils/scanFiles.js
@@ -53,16 +53,16 @@ let client = serviceKey
   : undefined
 
 const contentModerateFile = (file, callback) => {
-  if (!SUPPORTED_FILE_TYPES.includes(file[1].type)) {
+  if (!SUPPORTED_FILE_TYPES.includes(file.type)) {
     console.debug(
-      `Content Moderator File not scanned Azure image moderator doesn't support for file type ${file[1].type}`,
+      `Content Moderator File not scanned Azure image moderator doesn't support for file type ${file.type}`,
     )
-    file[1].adultClassificationScore =
-      'Could not scan - not a supported file type'
-    callback(null, file[1])
+    file.adultClassificationScore = 'Could not scan - not a supported file type'
+    callback(null, file)
     return
   }
-  var readStream = fs.createReadStream(file[1].path)
+  var readStream = fs.createReadStream(file.path)
+
   client.imageModeration.evaluateFileInput(readStream, {}, function (
     err,
     _result,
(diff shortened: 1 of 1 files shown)


sdzhangtao:
I know NodeJS is thread safe, but only safe against variable corruptions. This is a grey area that we will have to build skills on. In Java, we know so well.
But we have to remember that some global variables are still shared. in this case the queue object.

sdzhangtao:
Also, we are running in docker, and might run multiple instances with a scale set. But this rate limitator is only for one instance. To be honest I would pay the 2 dollars to avoid this complex. And we also know we don't need this code in prod

justinr86:
> I know NodeJS is thread safe, but only safe against variable corruptions. This is a grey area that we will have to build skills on. In Java, we know so well.
> But we have to remember that some global variables are still shared. in this case the queue object.

Yes, I believe this is something to discuss in the next JavaScript Deep Dive meeting. 

If I understand the issue correctly, we want to limit the number of calls to Content Moderator to 1 per second. Since the submission is dependent on data from Content Moderator, which is called once per file, this is not scalable. My assumption was that this was only a requirement while the application was in Beta testing.

sdzhangtao:
> I know NodeJS is thread safe, but only safe against variable corruptions. This is a grey area that we will have to build skills on. In Java, we know so well.
> But we have to remember that some global variables are still shared. in this case the queue object.

If the site has concurrent requests, they are all push into the queue. But the queue.drain would wait until all tasks done before await returns

f2/src/utils/scanFiles.js
```diff
@@ -76,34 +78,59 @@ const contentModerateFile = (file, callback) => {
...
         console.warn(`Error in Content Moderator: ${error.stack} `)
       }
     }
-    callback(null, file[1])
+    callback(null, file)
   })
 }
 
+//Queue to call content moderator once per second.
+const queue = async.queue((task, callback) => {
```
sdzhangtao, line 98: This queue object would a global object. Accessible by all requests

f2/src/utils/scanFiles.js
```diff
@@ -76,34 +78,59 @@ const contentModerateFile = (file, callback) => {
...
-      Object.entries(data.evidence.files),
-      contentModerateFile,
-      function (err, _results) {
-        if (err) console.warn('Content Moderator Error:' + JSON.stringify(err))
-        finalCallback()
-      },
-    )
+  } else {
+    //Add files to queue
+    queue.push(data.evidence.files)
```
sdzhangtao, line 128: Each concurrent request would push all files into this queue

f2/src/utils/scanFiles.js
```diff
@@ -76,34 +78,59 @@ const contentModerateFile = (file, callback) => {
...
+
+    await queue.drain()
```
sdzhangtao, line 130: Would this cause everybody to wait for everybody?

f2/src/utils/scanFiles.js
justinr86: Yes, if we want the app to only call the server once per second we will need to use the same queue. Otherwise overlapping requests could cause requests less than one second apart.

f2/src/utils/scanFiles.js
justinr86: This will cause all requests to be processed before submission. My thinking was that since we are limiting the number of submissions this won't have much of an impact.

Overall this rate limit is not scalable. I guess my question is - Is this feature intended for production or is this simply needed to manage cost while Beta testing? If it is intended for production we need to discuss further.

I could write a call back for each request. This way the submissions won't be waiting on one another.

sdzhangtao: For the complexity this introduces, I am not even sure we should go ahead.
But this problem will haunt me. I can find a solid solution for java. But I find it hard to do in NodeJS.

justinr86:
Code change unnecessary.

PR #2138, opened 2020-07-28
Update How Did It Start

justinr86:
This is the first form to be redesigned using Formik, Yup, and React Bootstrap. 

NPM packages for Formik, Yup, React Bootstrap, and React Icons have been added to the project.

A validation schema has been created for this form. For other forms add a definition to this object. You will see in **validationSchema.js** (line 8 - 15) a nested object is used to define the schema for the How Did It Start form. To create a validation schema for another for simply add another nested object and export it using **Yup.reach()**. For more information on defining rules or other Yup functionality documentation can be found [here](https://github.com/jquense/yup).

Note that the validation for this form is for testing purposes only. There are no validation rules defined for this form in production so I created one to test/demonstrate.

New form components have been created using React Bootstrap. I used CSS to style these to achieve a similar look to existing components. This will be an area for improvement for sure, perhaps in the future someone with better CSS skills can create an issue for this. The checkbox, textarea, start and cancel button components have been created in this PR. We will need to create more components for other forms (i.e. select, input, radio buttons, etc). React Bootstrap documentation can be found [here](https://react-bootstrap.github.io/components/forms/).

Formik has been added as a new form library. The main benefit is the ability to pass the Yup validation schema to the form, Formik will take care of the rest. For this update I have passed custom components to the Field props (**HowDidItStartForm.js** lines 107 and 130). 

Additional styling needed to take place due to bootstrap's CSS overriding some of our components. This is why ButtonLink components have classes added to them. 

This is a relatively big change and will likely need some adjustment before it can be merged. Please review and test, I would like feedback regarding styling, convention, and quality. Once this is merged we can begin making similar changes to the rest of these forms.

HowDidItStart


sdzhangtao:
It feels like that every single field within the form will be re-rendered for any change that will change form values. This might be a performance issue. Especially when the form has many fields. Formik has FastField for this purpose. But we need to research more in this regard. Pay attention if the UI component takes time to initialize, ie, load states across the network etc.

sdzhangtao:
I actually don't like the nested validation schemas. Given any reasonable sized project, this validation schema file will be crowded and becomes the contention point. Encapsulation is better if we define the validation schema closer to the form. However the most basic/primitive validation schema  (such as email, phone, name etc) better be centralized and easily shared.

f2/src/components/formik/input/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const Input = ({ field, form, ...props }) => {
+  return (
+    <Row className="form-row">
+      <Form.Group controlId={props.id}>
+        <Form.Label>{props.label}</Form.Label>
+        <Form.Text className="input-help-text">{props.helpText}</Form.Text>
+        <Form.Control
+          className="input-field-control"
+          type={props.type}
```
sdzhangtao, line 13: We should include this. 
`          {...field}`

justinr86: Thanks for pushing this John.

f2/src/components/formik/radio/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
```
sdzhangtao, line 1: Can we merge Radio and Checkbox into one component and pass in a type prop? Those two components are mostly the same except the type.

justinr86: This was discussed with @ngosset and it was decided that these should remain separate in case different functionality is required in the future.

f2/src/utils/formik/validationSchema.js
```diff
@@ -0,0 +1,27 @@
+const Yup = require('yup')
+
```
sdzhangtao, line 2: As we discussed, let move all the regex or Yup schema object into a separate package and export them.
Then we can import required regex or Yup schema object here.

f2/src/utils/formik/validationSchema.js
```diff
@@ -0,0 +1,27 @@
...
+const phoneRegExp = /^(\+?\d{0,4})?\s?-?\s?(\(?\d{3}\)?)\s?-?\s?(\(?\d{3}\)?)\s?-?\s?(\(?\d{4}\)?)?$/
+const postalCodeRegex = /^([A-Za-z]\d[A-Za-z][-]?\d[A-Za-z]\d)/
+
+//Validation object, defines rules used to validation form data
+const validationSchema = Yup.object().shape({
```
sdzhangtao, line 7: I think it's a better idea to co-locate validation schema with different forms. To avoid this js file becoming the contention point

f2/src/components/formik/button/index.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const CancelButton = (props) => {
+  return (
+    <Route
+      render={({ history }) => (
+        <Button
+          type="button"
+          className="button cancel-button"
+          to="/confirmCancel"
```
sdzhangtao, line 25: React-bootstrap probably doesn't know React Route well. I am wondering what's the use of to here.
Since you have already onclick to do a history push

f2/src/components/formik/button/index.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const SkipButton = (props) => {
+  return (
+    <Route
+      render={({ history }) => (
+        <Button
+          type="button"
+          className="button skip-button"
+          to={props.to}
```
sdzhangtao, line 67: Same thing here

ying-chen-cds:
I tried how did it start page. I can check the checkbox, but can't uncheck it. And also, I can only check the checkbox when I put the mouse in certain area of the square, that's confusing.

justinr86:
> I tried how did it start page. I can check the checkbox, but can't uncheck it. And also, I can only check the checkbox when I put the mouse in certain area of the square, that's confusing.

Please access components via the test form only. Once these new components have been approved howDidItStart will be updated.

f2/src/utils/formik/validationSchema.js
justinr86: Good point, I will move schema definition to the form. However, this validation logic is only for demonstration purpose since there is no existing validation rules for this page.

f2/src/components/formik/button/index.js
justinr86: I just copied the pops from the Chakra-UI cancel button. You are right, I tested without the **to** prop and it works. I have updated the component.

f2/src/components/formik/button/index.js
justinr86: Agreed.

f2/src/utils/formik/validationSchema.js
justinr86: Yup validation logic is now defined within the form. It works as expected.

f2/src/components/formik/datePicker/index.js
```diff
@@ -0,0 +1,88 @@
...
+const DateEntry = ({ field, form, ...props }) => {
+  return (
+    <Col sm="auto" className="date-picker">
+      <Form.Group controlId={props.id}>
+        <Form.Label className="date-label">
+          {<Trans id="whenDidItStart.startDay" />}
+        </Form.Label>
+        <Form.Control
+          className="day-month date-input"
+          type="input"
```
sdzhangtao, line 15: Is there an 'input' type in HTML? You might mean 'text'
type='text'

Based on react-bootstrap API, as by default is input, then we can specify type as text, email, password etc

f2/src/components/formik/datePicker/index.js
```diff
@@ -0,0 +1,88 @@
...
+import { Field } from 'formik'
+
+const DateEntry = ({ field, form, ...props }) => {
+  return (
+    <Col sm="auto" className="date-picker">
+      <Form.Group controlId={props.id}>
+        <Form.Label className="date-label">
+          {<Trans id="whenDidItStart.startDay" />}
+        </Form.Label>
+        <Form.Control
```
sdzhangtao, line 13: Can we reuse our new Input component for all three input fields, date, month and year?

f2/src/utils/formik/validationSchema.js
justinr86: Regex is defined in a separate module, Yup schema is defined in the form now.

f2/src/components/formik/datePicker/index.js
justinr86: Corrected.

f2/src/components/formik/datePicker/index.js
justinr86: Done, DateEntry component is reused for each Field.

DianeLiu2019:
test it , all good except some css styles little bit different from the current one. we can change later.

PR #2235, opened 2020-08-20
Rewrite what happened form

### LINKED ISSUE #2157: Update What Happened Form
Update WhatHappenedForm.js to use Formik, Yup, and Bootstrap as documented in issue #2136. These changes have been made to HowDidItStartForm.js, you can them here - #2138.  

Textarea component has already been created. This one should be pretty straight forward.

ying-chen-cds:
Fixes #2157 

# Description

> Refactor whatHappenedForm

# Any new packages installed?

> Give details

# Required followup work

# Checklist:


f2/src/forms/WhatHappenedForm.js
```diff
@@ -2,23 +2,14 @@
 import PropTypes from 'prop-types'
 import { jsx } from '@emotion/core'
 import { Trans } from '@lingui/macro'
-import { Form } from 'react-final-form'
-import { Field } from '../components/Field'
 import { TextArea } from '../components/text-area'
```
justinr86, line 5: Please use the textarea component from f2\src\components\formik\textArea\index.js

ying-chen-cds: Fixed

f2/src/components/formik/textArea/index.js
```diff
@@ -13,6 +13,7 @@ export const TextArea = ({ field, form, ...props }) => {
         className={
           'open-text-entry' + (largeTextArea ? ' large-text-area' : '')
         }
+        {...props}
```
justinr86, line 16: Why are we modifying the TextArea component? What properties are we trying to pass to Form.Control that are not present?

f2/src/forms/WhatHappenedForm.js
```diff
@@ -28,45 +19,61 @@ export const WhatHappenedForm = (props) => {
...
+          <Container>
+            <Row className="form-question">
+              <Row className="form-label">
+                {formLabel}
+              </Row>
+              <Row className="form-helper-text">
+                {formHelpText}
+                {formHelpText2 ? (
+                <p>
+                  <br />
```
justinr86, line 50: We could probably replace this with Jonh's <P> component from f2\src\components\formik\paragraph\index.js.

f2/src/forms/WhatHappenedForm.js
```diff
@@ -28,45 +19,61 @@ export const WhatHappenedForm = (props) => {
...
+                ) : null}
+              </Row>
+            </Row>
+            <Row className="form-section">
+              <Field
+                name="whatHappened"
+                component={TextArea}
+                onChange={handleChange}
+                onBlur={handleBlur}
+                rows="12"
```
justinr86, line 62: The rows property does not take effect. Rows is an HTML attribute and will not be rendered into a proper CSS rule. To get a large TextArea pass the prop size="large".

justinr86:
The text at the top of the form is not aligned, the first paragraph overflows. Also the text area is small, add the property size="large" to create a larger text area.

whatHappened

f2/src/components/formik/textArea/index.js
ying-chen-cds: "rows", to adjust the height of the textArea. I know you add 'large-text-area', but that is a fixed height (300px). It is more flexible to pass props so that you can define any height for the textArea by "rows", and potentially in the future, we could pass in other props for the textArea to make this component more flexible.

f2/src/forms/WhatHappenedForm.js
ying-chen-cds: It will if you pass it to Form.Control. I tested.
I know you add 'large-text-area', but that is a fixed height (300px). It is more flexible to pass props so that you can define any height for the textArea by "rows", and potentially in the future, we could pass in other props for the textArea to make this component more flexible.

f2/src/forms/WhatHappenedForm.js
ying-chen-cds: This is from the old code (maybe yours:)) and it works well. May I ask the reason you want to replace it?

ying-chen-cds:
Weird. The textArea works properly on my local. Don't know why it looks like this here. I'll check. Thanks!
whatHappened

justinr86:
> Weird. The textArea works properly on my local. Don't know why it looks like this here. I'll check. Thanks!
> whatHappened

This seems to only be a problem when I view it with Heroku using the "View Deployment" button. I should have tested locally.

f2/src/forms/WhatHappenedForm.js
justinr86: 300px was the height used previously with Chakra-UI components. 

Since the prop is being passed after all please remove the largeTextArea code from TextArea component. In the future we will likely use Emotion to style the components so this class was simply a stop-gap anyway.

f2/src/forms/WhatHappenedForm.js
justinr86: No need to replace, just thought it would be a good time to use the <P> component with Emotion.styled. It's your call.

ying-chen-cds:
> The text at the top of the form is not aligned, the first paragraph overflows. Also the text area is small, add the property size="large" to create a larger text area.
> 
> whatHappened

I used "form-helper-text" you created, that section of the code is same as HowDidItStartForm.  Not sure why the first paragraph overflows?

f2/src/forms/WhatHappenedForm.js
ying-chen-cds: OK, I removed largeTextArea as you suggested.

justinr86:
Looks like adding all of the props to TextArea component does have unintended consequences. 

The conditional field style has been passed to the TextArea in the HowDidItStart form.
textArea

Also, by examining the element you can see that other attributes are being injected as well.
textArea2

ying-chen-cds:
> Looks like adding all of the props to TextArea component does have unintended consequences.
> 
> The conditional field style has been passed to the TextArea in the HowDidItStart form.
> textArea
> 
> Also, by examining the element you can see that other attributes are being injected as well.
> textArea2

Hi Justin, 
I don't see that effect. It even works with Herouku by "View deployment" now. Could that be just an environment issue?
whatHappened

justinr86 (approved):

PR #2269, opened 2020-09-03
Update when did it happen

justinr86:
Update validation for When Did It Happen form.

More discussion will be required and some error messages are hard coded but this is a good start. Once we agree on the validation approach we can have the messages translated/approved.


f2/src/forms/WhenDidItHappenForm.js
```diff
@@ -76,68 +46,53 @@ export const WhenDidItHappenForm = (props) => {
...
-    return errors
   }
 
   return (
     <React.Fragment>
       <Formik
         initialValues={whenDidItHappen}
-        initialStatus={{ errors: '' }}
-        validationSchema={whenDidItHappenFormSchema()}
         validateOnChange={false}
```
sdzhangtao, line 67: Do we still need validateOnChange={false}

f2/src/forms/WhenDidItHappenForm.js
```diff
@@ -76,68 +46,53 @@ export const WhenDidItHappenForm = (props) => {
...
         }}
       >
-        {({ handleSubmit, handleChange, handleBlur, status, errors }) => (
+        {({ handleSubmit, handleChange, handleBlur, submitCount, errors }) => (
           <Form onSubmit={handleSubmit}>
-            <FormikListener />
             <Container>
               <Row className="form-question">
-                <Row className="form-label">
+                {Object.keys(errors).length > 0 && (
```
sdzhangtao, line 85: Is it a good idea to check errors.size inside ErrorSummary component? Instead of wrapping the check outside?

f2/src/forms/WhenDidItHappenFormSchema.js
```diff
@@ -1,55 +1,236 @@
...
-  happenedOnceYear: Yup.number().when(
-    'incidentFrequency',
-    validationSelection('once', dateSchema.YEAR),
-  ),
-  description: Yup.string(),
-})
-
-export const whenDidItHappenFormSchema = () => {
-  return whenDidItHappenForm
+import React from 'react'
```
sdzhangtao, line 1: Should we rename this file to WhenDidItHappenFormValidator?

f2/src/forms/WhenDidItHappenForm.js
```diff
@@ -76,68 +46,53 @@ export const WhenDidItHappenForm = (props) => {
...
-          if (Object.keys(errors).length === 0) {
+        validate={(values) => {
+          return realTimeValidation(values)
+        }}
+        onSubmit={async (values, { setErrors }) => {
+          const errors = onSubmitValidation(values)
+          if (errors.fields) {
+            setErrors(errors.fields)
+          } else {
+            await clearData(values)
```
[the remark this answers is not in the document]
justinr86: Need await, no. I just put it in to be safe. Considering the simple nature of clearData it could be removed.

f2/src/forms/WhenDidItHappenFormSchema.js
justinr86: We could rename. The form schema is defined in this file though. Would it be more appropriate to have two files (schema and validator)?

f2/src/forms/WhenDidItHappenForm.js
justinr86: You're right, this option is no longer needed.

f2/src/forms/WhenDidItHappenFormSchema.js
sdzhangtao: Yes, two files are perfect. Schema is a naming convention resevered for Yup

f2/src/forms/WhenDidItHappenForm.js
justinr86: Due to use of hooks in the component this would require further refactoring. Not a bad idea though.

ying-chen-cds:
I like the way CDS did, i.e., put the error both at the top of the page and the top of the field that needs to be corrected. When you click the error, it brings you to to that field, and you see an error message right there make you very confident that it is the field you need to correct. The new code brings you to that field, but there is no error indicator there (or even make the field red would be more intuitive).

ying-chen-cds:
There seems some bugs in the code. If you toggle the radio buttons with the error message, you'll see some weird behavior. Here are the steps to reproduce one scenario:
1. Go to when did it happen page
2. Choose "Once" and click "Continue", you'll see an error message at the top of the form, that's good.
3. Now click "More than once", and the error message at the top of the form is still there, is that right?
4. Click the link in the error message, the error message disappeared. I don't think that's correct.

justinr86:
> There seems some bugs in the code. If you toggle the radio buttons with the error message, you'll see some weird behavior. Here are the steps to reproduce one scenario:
> 
> 1. Go to when did it happen page
> 2. Choose "Once" and click "Continue", you'll see an error message at the top of the form, that's good.
> 3. Now click "More than once", and the error message at the top of the form is still there, is that right?
> 4. Click the link in the error message, the error message disappeared. I don't think that's correct.

Thanks for finding this. Fixed.

f2/src/components/formik/datePicker/index.js
```diff
@@ -39,34 +39,31 @@ export const DatePicker = ({ field, form, ...props }) => {
...
-          <Error>{form.status.errors[field.name + 'Error']}</Error>
-        )}
-      </Row>
-
       <Row className="date-group">
         <Field
           name={field.name + 'Day'}
           component={DateEntry}
           id={props.id + 'Day'}
           value={props.day}
```
sdzhangtao, line 47: Do we really need this line of code?
value={props.day}
It applies to month and year as well

justinr86: Tested this and it is not necessary. Update has been pushed.

sdzhangtao (approved):
Looks really good.

PR #2295, opened 2020-09-18
What was affected navigation

### LINKED ISSUE #1377: Suggestion: Should it direct to required page instead of summary page?
## Summary
The user fills out the report with affecting on his personal information and finance. When it goes to the review page, he just remembers that it can affect his business as well. He clicks on the "Edit" button at section "What could be affected?", chooses "Business assets" and presses <Continue> button. 

## Design Detail

Currently, it will go back to the summary page and if the user knows, he has to press <edit> button at section "Business" and enter more info. The important info that helps to investigate can be missing if the user does not know or forget to edit that session?
It will be better if the required pages will be displayed (if required) instead of summary page 

> Include an image of any old vs new design changes if possible

> include links to figma or other design designs if possible

## Unresolved questions

justinr86:
Fixes #1377 

# Description

Updates to page navigation regarding What Was Affected pages. Modifying What Was Affected from the confirmation page will present the pages for new corresponding selections. 

# Any new packages installed?

N/A

# Required followup work

N/A

# Checklist:


f2/src/BusinessPage.js
```diff
@@ -35,16 +39,28 @@ export const BusinessPage = () => {
                 </P>
               </Stack>
               <BusinessInfoForm
-                onSubmit={(data) => {
+                onSubmit={async (data) => {
+                  if (doneForms && !pages.editOptions) {
+                    pages.nextPage = whatWasAffectedPages.CONFIRMATION
+                  } else {
+                    pages.nextPage = await nextPage(pages)
```
sdzhangtao, line 46: Is await necessary here? If you put await in front of a synchronous function, in my impression, the function will return a promise, even though the promise is immediately resolvable. I need to double check this.
If it's not required, I prefer remove it.

sdzhangtao (commented):
Feels like every page needs this piece of logic, should we adopt a solution such as RedirectRoute component, putting it in one central location.
This is a premature thought, I will look into further.

f2/src/BusinessPage.js
justinr86: You're right await is not necessary. I will push an update without await.

justinr86:
> Feels like every page needs this piece of logic, should we adopt a solution such as RedirectRoute component, putting it in one central location.
> This is a premature thought, I will look into further.

This is an interesting idea. To be honest it bothered me to have duplicate logic. I will look into a component that contains this that can be exported once.

justinr86:
> > Feels like every page needs this piece of logic, should we adopt a solution such as RedirectRoute component, putting it in one central location.
> > This is a premature thought, I will look into further.
> 
> This is an interesting idea. To be honest it bothered me to have duplicate logic. I will look into a component that contains this that can be exported once.

@sdzhangtao  Update pushed to use function with callback to remove duplicate logic. Let me know what you think.

f2/src/utils/nextWhatWasAffectedUrl.js
```diff
@@ -1,16 +1,114 @@
...
+    key: '',
+    url: 'whathappened',
+  },
+  CONFIRMATION: {
+    key: '',
+    url: 'confirmation',
+  },
+}
+
+export const pages = {
```
sdzhangtao, line 40: This would be much easier to understand if we call it whatWasAffectedNavState or some sort. pages creates lots of confusion when I read through the code especially at those selected pages.

f2/src/utils/nextWhatWasAffectedUrl.js
```diff
@@ -1,16 +1,114 @@
...
+  firstPage: whatWasAffectedPages.FIRST_PAGE,
```
sdzhangtao, line 41: Maybe it's best to take this out and leave it as a constant

f2/src/utils/nextWhatWasAffectedUrl.js
```diff
@@ -1,16 +1,114 @@
...
+  lastPage: whatWasAffectedPages.LAST_PAGE,
```
sdzhangtao, line 42: Maybe it's best to take this out and leave it as a constant

f2/src/utils/nextWhatWasAffectedUrl.js
sdzhangtao, line 41: What's the purpose of firstPage here? It seems never being used anywhere

f2/src/utils/nextWhatWasAffectedUrl.js
sdzhangtao, line 42: The lastPage could be removed, it only holds two states, either Confirmation page or what happened page. It's initialized to be what happened page, but set to confirmation page once doneForms being true.
This would simply pages object, also you could remove 
  if (doneForms) {
    pages.lastPage = whatWasAffectedPages.CONFIRMATION
  }
from all affected pages.
And move the logic into nextWhatWasAffectedUrl
export const nextPage = (doneForms, navObject) => {
  try {
    const selectedOptions = navObject.affectedOptions
    const lastSelection = selectedOptions.slice(-1)[0]
    let nextPage
    let index

    //If there are no more pages to display proceed to What Happened or Confirmation.
    if (
      navObject.currentPage.key === lastSelection ||
      selectedOptions.length === 0
    ) {
      if (doneForms)  return whatWasAffectedPages.CONFIRMATION
      return whatWasAffectedPages.LAST_PAGE
    }

f2/src/InformationPage.js
```diff
@@ -37,15 +41,19 @@ export const InformationPage = () => {
 
               <InformationForm
                 onSubmit={(data) => {
+                  navigate(doneForms, pages, (updatedPages) => {
+                    dispatch({
```
sdzhangtao, line 45: I always feel that I am missing something in React. Do we really need to call dispatch? I understand that we have nav stage stored in global state. But the core value of dispatch is to re-render whoever access that value in other pages or other components in the same page. 
But in this case, we use following code immediately. It feels to me that we don't really need to dispatch since pages.nextPage.url is up to date.
history.push(pages.nextPage.url)

sdzhangtao:
> > > Feels like every page needs this piece of logic, should we adopt a solution such as RedirectRoute component, putting it in one central location.
> > > This is a premature thought, I will look into further.
> > 
> > 
> > This is an interesting idea. To be honest it bothered me to have duplicate logic. I will look into a component that contains this that can be exported once.
> 
> @sdzhangtao Update pushed to use function with callback to remove duplicate logic. Let me know what you think.

It looks much much better now.

f2/src/InformationPage.js
sdzhangtao: Tell me if I miss some fundamentals. From time to time I feel so junior.

f2/src/utils/nextWhatWasAffectedUrl.js
justinr86: Object has been renamed.

f2/src/utils/nextWhatWasAffectedUrl.js
justinr86: This attribute has been removed.

f2/src/utils/nextWhatWasAffectedUrl.js
justinr86: This has been implemented.

f2/src/InformationPage.js
justinr86: You are correct. This was my misunderstanding - I thought we had to update state to reflect changes.

sdzhangtao (approved):

PR #2299, opened 2020-09-21
Fix spacing among checkbox

### LINKED ISSUE #2288: Cosmetic bug: Should we add more space between options in What were affected page?
## Summary

When the page is minimized, there is no space between Business and Other options (please see the below picture).  

image

## Steps to reproduce

## Unresolved questions

ying-chen-cds:
Fixes #2288 

# Description

>  Fix the problem When the page is minimized, there is no space between Business and Other options in what was affected page.

# Any new packages installed?

> Give details

#

# Checklist:


f2/src/theme/style.css
```diff
@@ -290,7 +290,6 @@ body {
 /*Checkboxes and Radio Buttons*/
 
 .custom-control {
-  height: 2.5rem;
```
justinr86, line 293: This breaks other styling. I agree this needs a CSS update but simply removing the height property from all elements that have the custom-control class is not the answer.

Also, since this needs an update it would be a good time to look into [Emotion](https://emotion.sh/docs/introduction) and [Styled-System](https://styled-system.com/getting-started) instead of CSS. This should make adjusting component styles easier in the future.

ying-chen-cds: May I ask which other styling does it break? I tested all the pages. Can you give me an example?

justinr86: This is a small example, but you can see the conditional element overlaps the checkbox. 
overflow

Also, here you see the spacing is inconsistent.
spacing

ying-chen-cds: The second example (mid-feedback), is still old CDS code (our new refactored code has not been merged yet). That's actually the correct way for arranging the space between checkboxes/radios (i.e., if there are more text, it should have more space, so the text of the first checkbox won't overlap with second one) 
Here is a test I did, if you put a long text (I just randomly put some text in for the first radio to make it longer)after a radio (same for checkbox)before removing height=2.5rem, it looks like this:
longtextBefore
After removing height=2.5rem, it looks like this. As you can see, this way is more responsive.
longtextAfter

ying-chen-cds: You are right about your first example, the conditional field is a bit closer to the checkbox than before, but you need to look carefully to find it. I took a screen shot for both before and after. But I can try to fix it if Hong complains:)
Before removing height=2.5rem, it looks like this:
conditionalFieldBefore
After removing height=2.5rem, it looks like this:
conditionalFieldAfter

ying-chen-cds: I know it's probably not the perfect solution (I am a beginner of CSS:)). But that does seem work better. Of course, if you have better solution, you are more than welcome:)

ying-chen-cds:
By the way, I tested it on the mobile, and it looks good.

f2/src/theme/style.css
justinr86: You are right, without the height an option with very long text is more responsive. However, the above change does indicate an unintended change to another element (even if it appears small). To restore the spacing of the conditional field could you try the following:

- Change the morgin-top of .conditional-field to 0.5rem

- Add padding-top: 0.25rem to .conditional-field

ying-chen-cds: Sure, I'll try that.

justinr86 (approved):

PR #2306, opened 2020-09-24
Button update

justinr86:
Update to button components to use Emotion and Styled-System instead of a single CSS file.


sdzhangtao:
The buttonTypes definition in theme.js may be better co-located with button/index.js. Since we are using CSS-in-JS, this style definition just like our styled, are similar in nature. All CSS applied to same components. No other components will share this definition, right?
Otherwise, our theme.js will have lots of definitions like this, unless we want to share this definition with other components.

f2/src/components/formik/button/theme.js
```diff
@@ -0,0 +1,47 @@
+export const buttonTypes = {
```
sdzhangtao, line 1: The buttonTypes definition in theme.js may be better co-located with button/index.js. Since we are using CSS-in-JS, this style definition just like our styled, are similar in nature. All CSS applied to same components. No other components will share this definition, right?
Otherwise, our theme.js will have lots of definitions like this, unless we want to share this definition with other components.

sdzhangtao (approved):

PR #2320, opened 2020-10-02
Resolve Storybook Module Not Found Error

justinr86:
Remove Winston Logger to prevent requirement of fs module. This logger is meant for server side logging and shouldn't be used in front end logic anyway.


justinr86:
Thanks to @dengcn and @DianeLiu2019, discussions helped push this forward.

dengcn (approved):
Verified this fix solved the error blocking the build

PR #2346, opened 2020-10-09
Update checkbox radio

justinr86:
Update to checkbox and radio button components to use Emotion and Styled-System instead of a single CSS file. Some CSS rules remain to override the Bootstrap rules that have a higher level of specificity.


f2/src/components/formik/conditionalField/index.js
```diff
@@ -0,0 +1,15 @@
...
+const FieldContainer = styled('div')`
+  border-left-width: 0.25rem;
+  border-left-color: rgb(174, 174, 174);
+  margin-left: 1.25rem;
+  padding-top: 1rem;
+  padding-left: 1.5rem;
+  width: -webkit-fill-available;
+`
+
+export const ConditionalField = (props) => {
```
sdzhangtao, line 13: Do we need this layer? Sorry my brain kind of short circuit here.
I feel we only need FieldContainer. If you just export FieldContainer as ConditionalField, it can achieve same thing.

f2/src/components/formik/checkbox/index.js
```diff
@@ -1,17 +1,51 @@
...
-import { Form, Row } from 'react-bootstrap'
+import styled from '@emotion/styled'
+import { Form } from 'react-bootstrap'
+import { FormRow } from '../row'
+import { ConditionalField } from '../conditionalField'
+import { checkboxRadio } from '../theme'
+import { cleanProps } from '../../../utils/cleanProps'
+
+const conditionalFieldStyle = (props) => {
+  if (props.checked && props.helpText) {
```
sdzhangtao, line 10: Just a note that the other two PRs, we use css helper function.

f2/src/components/formik/conditionalField/index.js
justinr86: Yes, you are right.

justinr86:
Latest updates combines the Radio and Checkbox components to prevent code duplication.

sdzhangtao (approved):

PR #2368, opened 2020-10-14
PR2189 enhanced phone information (International Number and validation) AND add phone extension

### LINKED ISSUE #2189: Phone number in contact info page should be validated
## Summary

Phone number in contact info page should be validated. Currently, it accepts the letters and less than 10 characters (full and fyi reports)

## Steps to reproduce

## Unresolved questions

dengcn:
Fixes #2189 

# Description
This is a feature delivery for followings
- Allow International Number for Phone Number, added validation accordingly
- Added phone extension and added validation accordingly
- Added new extension information to Summary
- Added new extension information to Final Analyst report

-regex
-yupschema
-contactform validation
-summary report
-Analyst report

# Any new packages installed?

> Give details

# Required followup work

# Checklist:

- [X ] I have looked at my code on GitHub and it all looks good (ex: no random commented out code or console.logs)


dengcn:
FYI:

The followings are examples of good international number formats which will get validation pass. This had been communicated with Norm, Stephanie and Hong.
 
'00116131122323' //=> true
'163454516131122323' //=> true
'++1-7134445+345' //=> true
'1.713.444.5345' //=> true
'+09-323-333-3332' //=> true
'853-12349543' //=> true
'8534444' //=> true
'(1)-(613)-(222-3333) //=> true

Checked with some international number specs, it requires Minimum 7 digits (some smallest country may fit into 7 digits).  
+, -, . , () are allowed, and any character other than these 5 and 10 digits will fail to pass validation.

Extension number is also validated, only [0] to [9] are allowed, otherwise error message will show

dengcn:
2189InternationalNumberExtension

f2/src/forms/ContactInfoForm.js
```diff
@@ -123,22 +123,55 @@ export const ContactInfoForm = (props) => {
...
                         />
                         {errors.phone && (
-                          <P
-                            color="#dc3545"
-                            fontSize="1.25rem"
-                            marginBottom="0.5rem"
-                          >
-                            {phone.errorMessage}
-                          </P>
+                          <Container>
```
justinr86, line 126: Why are we adding containers everywhere?

f2/src/forms/ContactInfoForm.js
```diff
@@ -123,22 +123,55 @@ export const ContactInfoForm = (props) => {
...
+                              label={<Trans id="contactinfoPage.phoneNumber" />}
+                              component={Input}
+                              onBlur={handleBlur}
+                              onChange={handleChange}
+                              id="phone"
+                            />
+                          </Col>
+                          <Col>
+                            <Field
+                              name="extension"
```
justinr86, line 164: Attributes for this field are defined in ContactInfoFormSchema.js, why aren't they being used here?

f2/src/utils/regex.js
```diff
@@ -2,5 +2,6 @@ export const regexDef = () => {
   return {
     phoneRegExp: /^(\+?\d{0,4})?\s?-?\s?(\(?\d{3}\)?)\s?-?\s?(\(?\d{3}\)?)\s?-?\s?(\(?\d{4}\)?)?$/,
     postalCodeRegex: /^[ABCEGHJKLMNPRSTVXY]\d[ABCEGHJKLMNPRSTVWXYZ]( )?\d[ABCEGHJKLMNPRSTVWXYZ]\d$/i,
+    isValidInternationalPhonenumber: /^\d{7,}$/,
```
justinr86, line 5: Is this simply checking for a 7 digit (or greater) number?

f2/src/utils/yupSchema.js
```diff
@@ -5,10 +5,13 @@ import { Trans } from '@lingui/macro'
 
 export const yupSchema = () => {
   return {
-    phoneSchema: Yup.string().matches(regexDef().phoneRegExp, {
-      excludeEmptyString: true,
-      message: 'Please enter a valid phone number',
-    }),
+    phoneSchema: Yup.string()
+      .transform((value) => value.replace(/[\s()+-.]|ext/gi, ''))
```
justinr86, line 9: Why are we removing the phone number regex?

justinr86: Phone number validation is broken now, "++--" is accepted as a valid phone number.

f2/src/forms/ContactInfoForm.js
justinr86: This field looks too big, do we want to reduce it's size?

f2/src/utils/regex.js
dengcn: This is what valid international number is required.

f2/src/utils/yupSchema.js
dengcn: As discussed in team stand-up, we need to populate this new international number rule to the other place where we use phone validation.  (How did incident start?)  Since business side has such needs where many scam/crime calls show international number. Or what could do is keep the old phone schema in the code for future use, while change the name of new international number to another name.  But from clean code perspective, we can just comment out phoneSchema (old), and add comments on new one if we wish to keep old validation for reference

justinr86: It is fine if we decide to modify what is valid phone number but this validation is broken. If we need to validate any type of phone number maybe it would be easier to use a node package like [phone](https://www.npmjs.com/package/phone).

f2/src/forms/ContactInfoForm.js
dengcn: That is true, I will reduce the length of extension field

f2/src/forms/ContactInfoForm.js
dengcn: Advice taken, I will remove extra Container tag

f2/src/utils/regex.js
justinr86: If we are validating international phone numbers shouldn't we use E.164 format instead of just check if it is 7 or more digits?

f2/src/utils/yupSchema.js
dengcn: Phone validation is not broken, this PR just lets Phone number in Contact form and (How did incident start) form follow the same rule that allows more loose validation since we allow International Number. The initiative for this request is to lower the validation bar when user types in phone number (Now we states international number to be allowed), so that we can reduce the not-completion rate when user fills the form.  For international number, Min. is 7, while E164 defines Max number either 12 or 15, it will add more complexity which may ends up adding (pick Country Code drop down menu, save a large number of valid country code etc). From what I heard from business team is to lower the validation bar for now until they have strong desire to let user to confirm country code.

justinr86: If this is the desired outcome why validate at all?
phone

dengcn: I just added one validation to limit the Max number of valid digits to 15

dengcn: "++++" needs to be fixed.

dengcn: Issue of "++++" like allowed charactors Only (with no digits) pass validation has been resolved in the new  commit

justinr86: Validation still needs work. What is the number of max digits - 15 for phone number, 10 for extension? See image below.
phone

dengcn: As for Extension, I added limitation to allowing up to 8 digits o valid number [0-9], it is already in latest push and commit.
As for Phone Number, as discussed in standup, we had better not put too much strict rule on International Number which will largely increase non-completion rate.  This app is NOT any business web site which involves financial transaction, not other government web site like CRA which has to enforce much stricter validation rule on any input from user since this involve real money and session is not URGENT.  Our app is more like 911 APP or work flow,  heard many stories of over strict/over engineered rule or validation may put user into desperate or live/dead situation or even revealing such experience on public media or news.

f2/src/utils/yupSchema.js
```diff
@@ -17,10 +22,16 @@ export const yupSchema = () => {
       message: <Trans id="locationInfoForm.Warning" />,
     }),
 
+    phoneExtensionSchema: Yup.string().matches(regexDef().phoneExtensionRegex, {
+      message: <Trans id="contactinfoForm.email.warning" />,
```
justinr86, line 26: Where is phoneExtensionRegex defined?

f2/src/utils/yupSchema.js
```diff
@@ -17,10 +22,16 @@ export const yupSchema = () => {
...
+    }),
+
     dateSchema: {
       DAY: Yup.number().min(1).max(31),
       MONTH: Yup.number().min(1).max(12),
       YEAR: Yup.number().min(1000).max(9999), //Enforce 4 digit entry
     },
   }
 }
+
```
justinr86, line 36: Should remove commented out code.

f2/src/utils/yupSchema.js
dengcn: Sorry  forget including/adding regex file, will add soon

f2/src/utils/yupSchema.js
justinr86: Extension does allow more than 8 characters.
extension

dengcn: One regex was not included in previous commit, now everything is in git

image

f2/src/utils/yupSchema.js
dengcn: removed comment left by others in master, I did not add this comment

dengcn:
From discussion with Stephanie and others in General section of TEAM, Phone validation is updated to ensure valid digits of phone should be more than or equal to 10 digits to ensure area code is captured. Max number is still 15 digits for international number use. Such code changes had been committed to GIT, and all test pass.

justinr86 (approved):

PR #2374, opened 2020-10-16
Update paragraph

justinr86:
Update to p components to use Emotion and Styled-System instead of a single CSS file. This update adds an ErrorText component to make displaying error messages easier.


sdzhangtao:
Can you update the home page to remove the test page? I think this one is ready to merge

sdzhangtao (approved):

PR #2387, opened 2020-10-20
Warning Before Go Back

### LINKED ISSUE #2345: Warning When Navigating Away From Page
## Suggestion / question

While filling out a form, if the user uses the back button of the browser or the BackButton component to navigate away from the page their current entries will be lost. It might be a good idea to display a warning before leaving the page. This would be especially useful on the What Happened page where a user could have entered multiple paragraphs. To display this message we could use a modal which was introduced when the Evidence form was refactored.

## Use case

When a user spends time entering information about their incident, if they navigate way from the page this information will be lost. This could cause frustration or confusion and will likely lead to user drop off.

justinr86:
Fixes #2345 

# Description

Add warning when user navigates away after entering information. There are two scenarios that will display a warning.

### Go Back

If the user enters information in a form and goes back to the previous page a modal will display a warning message. The user will have the option to proceed to the previous page (OK), or remain on the page (Cancel).
modal_en

### Leaving the App

If the user started a report but navigates away from the app by entering a URL in the address bar a warning message will be displayed. This will be the browser default message that uses it's language setting, not the currently selected language. 
browserMessage

# Any new packages installed?

N/A

# Required followup work

N/A

# Checklist:


DianeLiu2019 (commented):
tested all the page. when going back, the error message don't show up in  anaonymous page and evidence page

f2/src/components/formik/warningModal/index.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  const cancel = () => {
+    setShowModal(false)
+  }
+
+  const shouldPrompt = (dirty, isSubmitting) => {
+    const alteredForm = dirty && !isSubmitting
+
+    console.log(alteredForm && !confirmNav)
```
DianeLiu2019, line 25: please don't forget to remove the " console.log " line before merge if you want

justinr86: Removed.

justinr86:
> tested all the page. when going back, the error message don't show up in anaonymous page and evidence page

Warning now appears on this page.

DianeLiu2019 (approved):
LGTM

PR #2401, opened 2020-10-21
Fix incorrect self harm alarm warning on server while no self harm is detected

### LINKED ISSUE #2333: Self Harm Words Incorrectly indicated in Server Console.
## Summary

When a report is submitted the we scan the contents for words that indicate the user is at risk for self harm. Any matches are flagged in the analyst report and printed in the server console. When no self harm words are found we still print an empty list in the console. 

sfwords

This may lead to confusion in the future and this should be removed. It seems the issue is the self harm word scan returns an array of matching words. The result is evaluated and if true the results are printed. I believe that originally the return type was String so this would have worked ("" is falsy, [] is truthy). The simple fix is to update the evaluation.

## Steps to reproduce

Complete a report without any self harm words. Check server console after submitting.

## Unresolved questions

N/A

dengcn:
Fixes #2333 (issue)

# Description
This is to fix self harm alarm warning incorrectly logged on server when no self harm is detected

# Any new packages installed?

> Give details
How to test:
1) Use harmful words in text description inputs, and submit report, check server side log, you should see something like:
"Self harm words detected: suicide,kill,knife,depressed" 

2) Do not use any harmful words in text description inputs, and submit report, check server side log,  you should not see any thing like
"Self harm words detected"

# Required followup work

# Checklist:

- [X ] I have looked at my code on GitHub and it all looks good (ex: no random commented out code or console.logs)


f2/src/utils/getData.js
```diff
@@ -51,7 +51,12 @@ async function getData(data, files) {
   data.evidence.files = filesToJson
 
   const selfHarmWords = selfHarmWordsScan(data)
-  if (selfHarmWords) {
+
+  if (
+    selfHarmWords !== null &&
```
justinr86, line 56: Why are we converting an array to String here? I think a better approach might be something like this:

`selfHarmWords === undefined || selfHarmWords.length == 0`

If it does not exist or if there are no elements in the array it will return true.

dengcn: Because there is still possibility of one element added to array, and this element contains empty value or null value

justinr86: If that is a possibility it is a bug. We should address it in this PR.

dengcn: Bug not in this specific case, but I saw similar situation many times in other projects, so I take this much stricter condition to rule out such things.

justinr86: This condition is only used to write a message to the console. The array will be used in other parts of the app, like when creating the analyst email. If this function will return an array containing blank strings or null values it should be fixed. If not we shouldn't have this check.

This check doesn't consider JavaScript types. If something goes wrong the array will be `undefined` instead of `null`. Calling toString on `undefined `will cause a `TypeError`. From Mozilla:

"A variable that has not been assigned a value is of type undefined. A method or statement also returns undefined if the variable that is being evaluated does not have an assigned value. A function returns undefined if a value was not returned."

[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined)

dengcn: Null is removed in latest commit.

dengcn: String length check is also removed to get it pass review. while I still think, as a preventive measure, it is useful to have such check to prevent such possible scenario.

justinr86 (approved):

PR #2406, opened 2020-10-22
CSS-in-JS Revision

justinr86:
Update to React Bootstrap components to use a CSS-in-JS approach versus a single CSS file. We will likely need more work in the future to improve this but it is another step in the right direction.


f2/src/components/formik/alert/index.js
```diff
@@ -30,7 +29,7 @@ const Summary = styled(BaseAlert, {
 `
 
 const iconStyle = () =>
-  css`
+  `
```
sdzhangtao, line 32: I know all CSS are hardcoded here.  And we don't need css helper function. But do they produce the same CSS?

justinr86: Yes, the resulting components are styled in the same way.

f2/src/components/formik/button/index.js
```diff
@@ -74,6 +121,16 @@ const PaperClipIcon = styled(FiPaperclip)`
   margin-right: 0.5rem;
 `
 
+const ButtonContainer = styled(Container)`
+  margin-bottom: 2.5rem;
+  margin-top: 2.5rem;
+`
+
+const ButtonOffset = styled(Col)`
```
sdzhangtao, line 129: Or we should call this ButtonCo? To follow your naming convention.

sdzhangtao (approved):

PR #2412, opened 2020-10-23
client log to server and ErrorBoundary

sdzhangtao:
Fixes #(issue)
Client is not able to log infomation and error messages to server, they lost with the client's browser. When the app crashes, we don't have a trace.
Client logs are not associated to each other
server side logs are not associated to each other
It's also nicer to have an Error Page instead of a blank screen in case there are rendering error 

# Description
Client logs are associated with a new session id
At submit time, the session id will be passed to server
server logs are associated with existing report id
Use existing winston logger, but works at client using winston HTTP transport to log to server
Generate new session id
Use React Error Boundar
# Any new packages installed?
No
> Give details

# Required followup work
Will have to design a new Error Page.

# Checklist:


justinr86:
I like this feature, we should probably remove the console.log messages from Landing Page. 

The only problem I see is the error and info messages are difficult to read, but this could be addressed in a subsequent update.

sdzhangtao:
> I like this feature, we should probably remove the console.log messages from Landing Page.
> 
> The only problem I see is the error and info messages are difficult to read, but this could be addressed in a subsequent update.

I removed the console.log. That's easy.
Do you mean formatted json for easy reading? For logging to analytics, one liner is preferable than formatted I read.

justinr86:
> > I like this feature, we should probably remove the console.log messages from Landing Page.
> > The only problem I see is the error and info messages are difficult to read, but this could be addressed in a subsequent update.
> 
> I removed the console.log. That's easy.
> Do you mean formatted json for easy reading? For logging to analytics, one liner is preferable than formatted I read.

Yes, I was referring to formatted JSON for viewing in console. This makes sense.

justinr86 (approved):

Issue #2365, opened 2020-10-14
The next label in How were your money or finances affected? page should be changed 


lmcbhvu:
## Summary

The next label in How were your money or finances affected? is different with next label in other pages such as "How was your personal information affected?" and "How were your devices or accounts affected?".  Currently, it is "Next: What happened?" and the other pages is "Next: Details about what was affected" 

In prod version, it is "Next: Details about what was affected"  in "How were your money or finances affected?" page

## Steps to reproduce

> How exactly can the bug be reproduced? Be very specific.

## Unresolved questions

> Are there any related issues you consider out of scope for this issue that could be addressed in the future?

ngosset:
This is a duplicate issue of #1295

ying-chen-cds:
Fixed by PR2417

PR #2417, opened 2020-10-26
Conditional next text on money lost, personal info, devices and business pages

### LINKED ISSUE #1295: Conditional next text on "what could be affected" page and subsequent impact pages
## Summary

What comes next will depend on what checkboxes were selected. This is also the case on the subsequent pages if multiple checkboxes are selected.

## Design Detail

Is it possible to make the "Next: ____title____" dynamic depending which conditions are chosen?

For example: 
"Next: How were your money and finances affected?" if money is selected

Screen Shot 2020-02-06 at 10 57 40 AM

ying-chen-cds:
Fixes #1295 #2365 

# Description

> Conditional next text on money lost, personal info, devices and business pages

# Any new packages installed?

> Give details

# Required followup work

# Checklist:


f2/src/forms/MoneyLostInfoForm.js
```diff
@@ -69,6 +69,12 @@ export const MoneyLostInfoForm = (props) => {
           <Trans id="transactionDate.error.hasNoMonth" />
           <Trans id="transactionDate.error.hasNoDay" />
           <Trans id="moneyLostPage.financialTransactions" />
+          <Trans id="previousPageofMoneyLost.nextPage" />
```
justinr86, line 72: We should avoid doing this whenever possible.

ying-chen-cds: Ya, agree.  I tried to put it where it defined, but it didn't work. Do you know a better way to do it? Thanks!

f2/src/BusinessPage.js
```diff
@@ -18,6 +18,7 @@ export const BusinessPage = () => {
   const whatWasAffectedNavState = state.whatWasAffectedOptions
 
   whatWasAffectedNavState.currentPage = whatWasAffectedPages.BUSINESS
+  navigate(doneForms, whatWasAffectedNavState)
```
justinr86, line 21: If we are going to change where this is called we should change the name, `navigate` might cause some confusion. Perhaps `updateNavigation` would better describe the function.

f2/src/forms/MoneyLostInfoForm.js
justinr86: As far as I know this is a limitation of Lingui. We are declaring what message property to use outside of React. To avoid this we would have to set the label with React in scope. Since we have this type of thing everywhere this isn't a show stopper, just thought I would point it out. This type of thing would probably be better resolved in a separate issue.

f2/src/BusinessPage.js
ying-chen-cds: Changed.

justinr86 (approved):

PR #2428, opened 2020-10-27
Update DOM on Language Change

### LINKED ISSUE #1896: French accessibility
## Summary

1. The html lang attribute is only set to "en". French Voiceover cannot function properly for francophone users without the lang attribute set to "fr"
2. The page title is permanently set to "Report a scam"

## Steps to reproduce

1. Using voiceover in french, in the french version of the app
2. Navigate through the app. Some text will be read with the french voice, interactables will be read in english
3. The page title remains untranslated.

## Unresolved questions

Route paths also remain untranslated. Is there a way to create translation for those?

justinr86:
Fixes #1896

# Description

Update language and title of HTML page when language is changed.

# Any new packages installed?

N/A

# Required followup work

We should have someone who speaks French verify that the screen reader is saying messages in French properly.

# Checklist:


f2/src/i18n.config.js
```diff
@@ -13,6 +13,11 @@ export async function activate(locale) {
     catalog = await import(
       /* webpackChunkName: "i18n-[index]" */ `@lingui/loader!./locales/${locale}.json`
     )
+
+    const title = locale === 'fr' ? 'signaler une fraude' : 'Report a scam'
```
sdzhangtao: Hi Justin,

LGTM, 
But what do you think of useEffect? To do it in Toggler component of LocaleSwticher.js.

justinr86: Good idea, I will test and update if everything works.

justinr86: Tested useEffect, it works when used in the LocaleSwitcher component. Toggler has an inverted locale prop to display the other language - Français when in English, and English when in French.

sdzhangtao (approved):

